PR #2, opened 2019-09-06
RFC: implement an EEE based messaging protocol in Bee

Alex6323:
[Rendered](https://github.com/Alex6323/bee-rfcs/blob/bee_e3/text/0000-bee-e3/0000-bee-e3.md)

### DIFF
--- a/text/0000-bee-e3/0000-bee-e3.md
+++ b/text/0000-bee-e3/0000-bee-e3.md
@@ -1,0 +1,111 @@
++ Feature Name: bee-e3
++ Start Date: 2019-09-05
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#42](https://github.com/iotaledger/bee/issues/42)
+
+# Summary
+
+This feature introduces a *pub/sub messaging mechanism* for the Bee node framework that allows asynchronous communication across all node components without them being directly dependent on eachother, thereby allowing the system to become very modular and loosely coupled. It arrives as part of a dedicated crate named `bee-e3`, which will expand the current `bee` workspace.
+
+This feature implements a subset of the **EEE** specification introduced by the IOTA Foundation. The following are the most relevant:
+
+* **Effect**: effects represent the data that is sent between entities and environments. They abstract over the concrete data type used to encode a particular information. One can think of them as *messages*.
+* **Environment**: environments are namespaces that can receive effects from "affecting" entities and forward them - unmodified - to entities that have "joined" one. However, environments must not communicate with other environments directly. The name of an environment *must* be unique and *should* be meaningful with the intent to best describe its purpose.
+* **Entity**: entities are actors, that not only can receive/broadcast effects from/to one or many environments, but also mutate or even replace them by applying some custom logic defined by the programmer. However, entities must not communicate with other entities directly. Entities that haven't joined an environment, but are affecting at least one can be thought of as *source entities*, while those that have joined environments, but are not affecting any can be thought of as *sink entities*. They are the entry and the exit points to the messaging system. That for example will allow for Bee to send gossip utilizing several different network protocols, which is a necessity in the IoT.
+
+# Motivation
+
+This messaging system will bring the following benefits:
+* the node framework becomes highly decoupled by preventing entities from interacting directly with eachother, but instead having them to publish their results via an environment, which in turn arbitrary many entities can subscribe to without the entity even noticing or caring.
+* the node framework overall becomes more resilient to partial system failures due to the fact that is based on a pub/sub model, where subscribers only react to something being published they don't require to have control over.
+* the node framework becomes more easily extendable. That also comes from the fact that it is based on a pub/sub model. Arbitrary many entities can subscribe to an environment, without any additional effort to support that.
+
+# Detailed design
+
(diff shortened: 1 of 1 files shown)


SuperFluffy:
@wusyong comment on https://github.com/iotaledger/bee/issues/42#issuecomment-527732389

> Regarding Effect, how about make this a trait? We implement this trait to the type we want and we can also use trait bound and trait object in this way.

SuperFluffy:
@Alex6323 commented on https://github.com/iotaledger/bee/issues/42#issuecomment-528307618:

> We were having many discussions about how to best model Effect in previous chats in the Discord channel, which lead Lukas to the idea of typed supervisors (using generics). I can't repeat everything here, but there are several reasons why this idea would be unhandy atm to pursue. I'm not saying that this is not a valid solution, but since it's definitely more complex, I'ld say we start with something more simple like the enum solution.
>
> Regarding your idea: The way I see it is that an Effect is an abstraction over different datatypes (you want to be able to send floats, integers, bytes, you name it), but not an abstraction over behavior which traits allow you to model. Effects per se have no own behavior. They are just data.
>
> That's why you're basically limited to pursue one of three solutions:
>
> + using enums
> + using generics
> + using runtime reflection
>
> I am willing to try any of those possible solutions. I favor starting out simple (enum), see where the limitations are and then switch to another rather than shooting with canons right away.

lunfardo314:
## Slack discussion between @Alex6323 and @lunfardo314:


> **Evaldas Drąsutis 10:59 AM**
> I have one fundamental comment on the current EEE model. I post it here before posting to the github. We need some extension of EEE model for qubic dataflow. It would bring some small overhead to Bee and to API.
>
> There's a possibility not to do this at the Bee level EEE. Then we will need qubic-specific supervisors as entities to the Bee EEE (or some kind of tunnel between Bee EEE and qubic supervisor). This may even be better option, because that small supervisor would be optimized, specifically for qubic dataflow, which is good. However, I still present the idea here for discussion.
>
> Any entity instance can subscribe to many environments. Currently, effect doesn't bring any information about the environment it is coming from, so the entity is ignorant about it. The extension would be for entity to be able to know the ID of the environment the effect is coming from or the name `(tag, index)` of the subscription within this specific entity.
>
> One possibility:
> 
> ```rust
> e3::subscribe_environment(&calc_tps, "GOSSIP_IN_SIGNAL_FROM_NEIGHBOR_1", > "source_env_1");
> e3::subscribe_environment(&calc_tps, "GOSSIP_IN_SIGNAL_FROM_NEIGHBOR_2", "source_env_2");
>
> if effect.get_subscription_id() == "source_env_1" { ... }
> ```
>
> (probably not the best example)
>
> The above supports the idea of the effect as a trait with the method to access the originator, but there can be other implementation options as well. The reason why we need it for qubic is to enable real dataflow configurations with entity attachments. Actually, I don't see any normal alternative to it.
>
> Let's say we split dataflow processing in two paths which ends up in environments ENV1 and ENV2.
> Then we want to merge those two processing paths in self synchronised way in one function call: `fun(ENV1, ENV2)`.
>
> Some entity could be listening to ENV1 and ENV2, but as there's no way of knowing where effect is coming from, we can't write an entity (statefull, buffered) which waits for both values to come to trigger the function call. The simple solution above would solve the problem.
>
> Much less fundamental comment: the wording `the environment subscribes to the entity` doesn't sound good to me, because environments are passive things: variables, sites, names. They don't do anything except being attachment sites.
>
> I'd use `entity affects the environment` instead. This is just a wording of course 

> **Alexander Schmidt 2:34 PM**
> Re the wording: you're absolutely right. Environments are passive on the conceptual level. Of course on the implementation level they are "execution units" just like entities when receiving and broadcasting effects. When I saw the similarities on that level I thought to myself: Well, hey, they are not that much different, are they? I should let the API reflect that and help the user by making the API call look very familiar to how he would have to subscribe an entity to an environment.
> 
> Re 'effect doesn't bring any information about the environment it is coming from'. Glad you bring this up. I was thinking about putting that into the proposal as well, because actually, I already did that in reee. let me dig that up: 
> https://github.com/Alex6323/reee/blob/master/src/eee/entity.rs#L22
src/eee/entity.rs:22
> 
> ```rust
> pub trait Entity: Send {
> ```
>
> as you can see the "processor" of an effect also receives the name of the originating environment. I hope I understood everything correcltly.
>
> **Evaldas Drąsutis 2:46 PM**
> name of the environment as an argument is overkill imho. Besides, it leads to dependencies on concrete environments withn entity. We need decoupling and abstraction. Why not the following: one byte of information in each enum of the effect. The byte would be a subscription index within the entity. First subscription, second subscription etc. It has meaning for the attachments of the entity which joined the environment, not for the environment itself. Up to 256 subscriptions (joins) for each entity, more than enough and minimal overhead
>
> In most cases entities don’t need to distinguish between environments they join
>
> **Alexander Schmidt 2:49 PM**
> true
> I will think about your idea Evaldas.
>
> **Evaldas Drąsutis 2:50 PM**
> Thanks :thumbsup:
>
> **Alexander Schmidt 2:50 PM**
> would be great if we could add this discussion on Gh, so people might contribute.
>
> **Evaldas Drąsutis 2:52 PM**
> Btw, one qubic = one static attachment = one static configuration of supervisor. Qubics don’t neede dynamicaly add/delete environments etc. In the end supervisor is disposed altogether with the wubic. That is whay I am saying one supervisor per qubic is also good solution.
>Ok, i will add it to github later
>
> **Alexander Schmidt 2:53 PM**
>Okay, thanks.

SuperFluffy:
The team has decided to postpone this RFC for the time being, because EEE is in a very fluent exploratory stage at this moment, and not critical for the core functionality of bee. We will revisit it at a more appropriate time.

PR #3, opened 2019-09-06
RFC: Transaction type

samuel-rufi:
[Rendered](https://github.com/rufsam/bee-rfcs/blob/master/text/0000-bee-model-crate/0000-bee-model-crate.md)

Based on https://github.com/iotaledger/bee/issues/43

### DIFF
--- a/text/0000-bee-model-crate/0000-bee-model-crate.md
+++ b/text/0000-bee-model-crate/0000-bee-model-crate.md
@@ -20,30 +20,30 @@ This feature is responsible for the creation and interpretation of transactions.
 
 A transaction can be understood as a **tuple** of 15 elements. Essentially, it's an **ordered list (sequence)** of following elements/fields: 
 
-| Name  | Description | Size |
+| Name  | Description | Size in trytes |
 | ------------- | ------------- | ------------- |
-| **signature_or_message_fragment** | contains the signature of the transfer or user-defined message data | 2187 trytes |
-| **address**  | receiver (output) if value > 0, or sender (input) if value < 0  | 81 trytes |
-| **value**  | the transferred amount in IOTA  | 27 trytes |
-| **obsolete_tag**  | another arbitrary user-defined tag  | 27 trytes |
-| **timestamp**  | the time when the transaction was issued  | 9 trytes |
-| **current_index**  | the position of the transaction in its bundle  | 9 trytes |
-| **last_index**  | the index of the last transaction in the bundle  | 9 trytes |
-| **bundle**  | the hash of the bundle essence | 81 trytes |
-| **trunk**  | the hash of the first transaction referenced/approved | 81 trytes |
-| **branch**  | the hash of the second transaction referenced/approved | 81 trytes |
-| **tag**  | arbitrary user-defined value | 27 trytes |
-| **attachment_timestamp**  |  the timestamp for when Proof-of-Work is completed | 9 trytes |
-| **attachment_timestamp_lowerbound**  |  *not specified* | 9 trytes |
-| **attachment_timestamp_upperbound**  |  *not specified* | 9 trytes |
-| **nonce**  | is the Proof-of-Work nonce of the transaction | 27 trytes |
+| `signature_or_message_fragment` | contains the signature of the transfer or user-defined message data | 2187 trytes |
+| `address`  | receiver (output) if value > 0, or sender (input) if value < 0  | 81 trytes |
+| `value`  | the transferred amount in IOTA  | 27 trytes |
+| `obsolete_tag`  | another arbitrary user-defined tag  | 27 trytes |
+| `timestamp`  | the time when the transaction was issued  | 9 trytes |
+| `current_index` | the position of the transaction in its bundle  | 9 trytes |
+| `last_index`  | the index of the last transaction in the bundle  | 9 trytes |
+| `bundle`  | the hash of the bundle essence | 81 trytes |
+| `trunk`  | the hash of the first transaction referenced/approved | 81 trytes |
+| `branch`  | the hash of the second transaction referenced/approved | 81 trytes |
+| `tag`  | arbitrary user-defined value | 27 trytes |
+| `attachment_timestamp`  |  the timestamp for when Proof-of-Work is completed | 9 trytes |
+| `attachment_timestamp_lowerbound`  |  *not specified* | 9 trytes |
+| `attachment_timestamp_upperbound`  |  *not specified* | 9 trytes |
+| `nonce`  | is the Proof-of-Work nonce of the transaction | 27 trytes |
 
 As reflected in the table, each field is of constant length. Totally, one transaction consists of **2673 trytes** (**8019 trits**).
 
 Transactions are classified as unique. Each transaction is identified by its hash, the transaction hash, which includes all fields of a transaction.
 Note, the transaction hash is not part of the transaction.
 
-Once a transaction is built, it can be identified by its transaction hash. Manipulating a transaction afterwards would result in the hash no longer matching the transaction, which is more like creating a new transaction.
+Once a transaction is built, it can be identified by its transaction hash. Manipulating a transaction afterwards would result in the hash no longer matching the transaction, which essentially is like creating a new transaction.
 Therefore, transactions should **not be modifiable** after construction. Transaction fields should be only accessible by getter functions.
 
 ## Exposed Interface
@@ -125,7 +125,7 @@ impl TransactionBuilder {
 
     pub fn build(self) -> (Transaction, TranscationMetadata) {
 
-        Ok(Pow::compute(&self))
+        Pow::compute(&self)
 
     }
     
@@ -163,11 +163,10 @@ impl TransactionBuilder {
 The **from_bytes(bytes: &Vec)** is striking here. It serves as constructor and allows to build transactions from received bytes, e.g. from the network socket.
 It's one way to build transactions. Transactions can be built also from ground up via *build()* from **TransactionBuilder**.
 
-In comparison to the Transaction model, the TransactionBuilder can be manipulated as desired. It consists of public accessible fields.
-Once *build()* is called, the set fields will be validated. Validation happens via the *TransactionBuilderValidator*. It checks if fields don't contain any invalid characters. Besides that, if not the full size of fields (e.g. signature_or_message_fragment) is used, it will pad the empty space up.
+In comparison to Transaction, TransactionBuilder can be manipulated as desired with the help of setter methods.
 
-Once the validation of the transaction fields was successful, Proof-of-Work (PoW) will be next.
-**PoW will return a tuple (Transaction, TranscationMetadata).**
+In build() Proof-of-Work (PoW) will be called.
+**PoW will return the tuple (Transaction, TranscationMetadata).**
 
 ### TransactionMetadata
 
@@ -285,11 +284,9 @@ Without any bee-model crate, nodes can not exchange transactions. Therefore this
 
 # Rationale and alternatives
 
-- The distinction between Transaction and Transaction Builder as well as Bundle and Bundle Builder makes the code cleaner. Properties are clearly assigned to specific data objects and not mixed up.
+- The distinction between Transaction and Transaction Builder makes the code cleaner. Properties are clearly assigned to specific data objects and not mixed up.
 
-- The proposed crate interface is relatively intuitive. Completely different alternatives did not naturally come to mind.
-
-- The validation logic could be done in setter functions instead of the separate validator class.
+- The proposed crate interface seems relatively intuitive. Completely different alternatives did not naturally come to mind.
 
 - This kind of interface is relatively minimal and easily extended upon in a future iteration.
 
@@ -299,8 +296,6 @@ Without any bee-model crate, nodes can not exchange transactions. Therefore this
 
 - Where should the actual Transaction struct be initialized/returned? Current ideas are to put it in the build() of the TransactionBuilder or as it currently is in the compute() of Pow.
 
-- Where should Transaction fields get validated? One-time validation once build() is called by the separate *TransactionBuilderValidator* as it currently is, or use **setters** in the TransactionBuilder and validate in each setter separately?
+- Should we introduce a fluent API to the TransactionBuilder?
 
-- Should we use setters for the TransactionBuilder?
-
-- Introducing a fluent API to the TransactionBuilder?
+- Should we use Option<T> for the fields in TransactionBuilder? 
--- a/text/0000-bee-model-crate/0000-bee-model-crate.md
+++ b/text/0000-bee-model-crate/0000-bee-model-crate.md
@@ -19,21 +19,21 @@ Settlements or data transfers can be done with the help of these transactions. P
 
 A transaction consists of several fields (e.g. address, value, timestamp, tag). Each field is of static length. One transaction consists of 2673 trytes:
 
+- **signature_or_message_fragment** contains the signature of the transfer or user-defined message data = 2187 trytes
+- **address** receiver (output) if value > 0, or sender (input) if value < 0 = 81 trytes
+- **value** the transferred amount in IOTA  = 27 trytes
+- **obsolete_tag** another arbitrary user-defined tag = 27 trytes
+- **timestamp** the time when the transaction was issued = 9 trytes
+- **current_index** the position of the transaction in its bundle = 9 trytes
+- **last_index** the total number of transactions in the bundle = 9 trytes
+- **bundle hash** is the hash of the entire bundle = 81 trytes
 - **trunk hash** the hash of the first transaction referenced/approved = 81 trytes
 - **branch hash** the hash of the second transaction referenced/approved = 81 trytes
-- **signature_and_message_fragment** contains the signature of the transfer or user-defined message data = 2187 trytes
-- **value** the transferred amount in IOTA  = 27 trytes
-- **address** receiver (output) if value > 0, or sender (input) if value < 0 = 81 trytes
-- **timestamp** the time when the transaction was issued = 9 trytes
+- **nonce** is the Proof-of-Work nonce of the transaction = 27 trytes
+- **tag** arbitrary user-defined value = 27 trytes
 - **attachment_timestamp** the timestamp for when Proof-of-Work is completed = 9 trytes
 - **attachment_timestamp_lowerbound** is a slot for future use = 9 trytes
 - **attachment_timestamp_upperbound** is a slot for future use = 9 trytes
-- **tag** arbitrary user-defined value = 27 trytes
-- **obsolete_tag** another arbitrary user-defined tag = 27 trytes
-- **nonce** is the Proof-of-Work nonce of the transaction = 27 trytes
-- **bundle hash** is the hash of the entire bundle = 81 trytes
-- **current_index** the position of the transaction in its bundle = 9 trytes
-- **last_index** the total number of transactions in the bundle = 9 trytes
 
 A bundle is a collection of specific transactions. Bundles are required to bundle related information.
 For example, if not all data fits into one transaction, it has to be split across multiple transactions. An example would be payment settlements - they require the help of multiple transactions.
@@ -57,9 +57,9 @@ struct Transaction {
 
     signature_fragments: String,
     address: String,
-    value: i64,
+    value: u64,
     obsolete_tag: String,
-    timestamp: i64,
+    timestamp: u64,
     current_index: usize,
     last_index: usize,
     bundle_hash: String,
@@ -67,9 +67,9 @@ struct Transaction {
     branch: String,
     nonce: String,
     tag: String,
-    attachment_timestamp: i64,
-    attachment_timestamp_lower_bound: i64,
-    attachment_timestamp_upper_bound: i64,
+    attachment_timestamp: u64,
+    attachment_timestamp_lower_bound: u64,
+    attachment_timestamp_upper_bound: u64,
 
 }
 ```
@@ -96,58 +96,12 @@ impl Transaction {
         &self.value
     }
 
-    pub fn obsolete_tag(&self) -> &String {
-        &self.obsolete_tag
-    }
-
-    pub fn timestamp(&self) -> &i64 {
-        &self.timestamp
-    }
-
-    pub fn current_index(&self) -> &usize {
-        &self.current_index
-    }
-
-    pub fn last_index(&self) -> &usize {
-        &self.last_index
-    }
-
-    pub fn bundle_hash(&self) -> &String {
-        &self.bundle_hash
-    }
-
-    pub fn trunk(&self) -> &String {
-        &self.trunk
-    }
-
-    pub fn branch(&self) -> &String {
-        &self.branch
-    }
-
-    pub fn nonce(&self) -> &String {
-        &self.nonce
-    }
-
-    pub fn tag(&self) -> &String {
-        &self.tag
-    }
-
-    pub fn attachment_timestamp(&self) -> &i64 {
-        &self.attachment_timestamp
-    }
-
-    pub fn attachment_timestamp_lower_bound(&self) -> &i64 {
-        &self.attachment_timestamp_lower_bound
-    }
-
-    pub fn attachment_timestamp_upper_bound(&self) -> &i64 {
-        &self.attachment_timestamp_upper_bound
-    }
+    ...
 
 }
 ```
 
-The Transcation_Builder struct contains public accessible fields. 
+The TranscationBuilder struct contains public accessible fields. 
 The set values can be validated in the build() function which then returns the constructed transaction.
 Moreover, also the Proof-Of-Work will be called in the build() function.
 
@@ -329,6 +283,8 @@ and helps achieve correctness among the data objects. Properties are clearly ass
 
 - The validation logic could be done in a separate validator class, which then will be called in the build() function.
 
+- This kind of interface is relatively minimal and easily extended upon in a future iteration.
+
 # Unresolved questions
 
-- How is Proof-Of-Work acceleration handled?
+- How to handle deserialization of the incoming, encoded transaction?
--- a/text/0000-bee-model-crate/0000-bee-model-crate.md
+++ b/text/0000-bee-model-crate/0000-bee-model-crate.md
@@ -1,4 +1,4 @@
-+ Feature name: `transaction-module`
++ Feature name: `bee-transaction`
 + Start date: 2019-09-06
 + RFC PR: [iotaledger/bee-rfcs#3](https://github.com/iotaledger/bee-rfcs/pull/3)
 + Bee issue: [iotaledger/bee#43](https://github.com/iotaledger/bee/issues/43)
@@ -44,7 +44,7 @@ Transactions are classified as unique. Each transaction is identified by its has
 Note, the transaction hash is not part of the transaction.
 
 Once a transaction is built, it can be identified by its transaction hash. Manipulating a transaction afterwards would result in the hash no longer matching the transaction, which essentially is like creating a new transaction.
-Therefore, transactions should **not be modifiable** after construction. Transaction fields should be only accessible by getter functions.
+Therefore, transactions should **not be modifiable** after construction. Transaction fields should only be accessible by getter functions.
 
 ## Exposed Interface
 
@@ -71,6 +71,8 @@ struct Transaction {
     nonce: Nonce
 }
 ```
+
+As reflected above, **each field gets its own type**. This makes the code more readable and clean. Furthermore it makes sure allocations are valid. The validation happens in each type separately.
 
 The type implementation is defined as follows:
 
@@ -159,14 +161,13 @@ impl TransactionBuilder {
 
 }
 ```
-
-The **from_bytes(bytes: &Vec)** is striking here. It serves as constructor and allows to build transactions from received bytes, e.g. from the network socket.
-It's one way to build transactions. Transactions can be built also from ground up via *build()* from **TransactionBuilder**.
-
 In comparison to Transaction, TransactionBuilder can be manipulated as desired with the help of setter methods.
 
-In build() Proof-of-Work (PoW) will be called.
-**PoW will return the tuple (Transaction, TranscationMetadata).**
+The **from_bytes(bytes: &Vec)** is striking here. It serves as constructor and allows to build transactions from received bytes, e.g. from the network socket.
+It's one way to build transactions. Another way would be from ground up via setters and use **build()** from **TransactionBuilder**.
+
+Also **Pow::compute(&self)** in **build()** striking here. This function calls the needed Proof-of-Work (PoW) for the transaction.
+As reflected in the code, the **PoW will return the tuple (Transaction, TranscationMetadata).**
 
 ### TransactionMetadata
 
@@ -284,7 +285,7 @@ Without any bee-model crate, nodes can not exchange transactions. Therefore this
 
 # Rationale and alternatives
 
-- The distinction between Transaction and Transaction Builder makes the code cleaner. Properties are clearly assigned to specific data objects and not mixed up.
+- The distinction between Transaction and TransactionBuilder makes the code cleaner. Properties are clearly assigned to specific data objects and not mixed up.
 
 - The proposed crate interface seems relatively intuitive. Completely different alternatives did not naturally come to mind.
 
@@ -296,6 +297,8 @@ Without any bee-model crate, nodes can not exchange transactions. Therefore this
 
 - Where should the actual Transaction struct be initialized/returned? Current ideas are to put it in the build() of the TransactionBuilder or as it currently is in the compute() of Pow.
 
-- Should we introduce a fluent API to the TransactionBuilder?
+- Should we introduce a fluent API to the TransactionBuilder for more easy manipulation?
 
 - Should we use Option<T> for the fields in TransactionBuilder? 
+
+- Should we use a intermediary building type like **TransactionBuilder -> UnpowedTransaction -> Transaction**
--- a/text/0000-bee-model-crate/0000-bee-model-crate.md
+++ b/text/0000-bee-model-crate/0000-bee-model-crate.md
@@ -1,17 +1,21 @@
-+ Feature name: `bee-model-crate`
++ Feature name: `bee-transaction-crate`
 + Start date: 2019-09-06
 + RFC PR: [iotaledger/bee-rfcs#3](https://github.com/iotaledger/bee-rfcs/pull/3)
 + Bee issue: [iotaledger/bee#43](https://github.com/iotaledger/bee/issues/43)
 
 # Summary
 
-This feature is responsible for the creation and interpretation of transactions and bundles.
-
-# Motivation
-
 IOTA is a distributed ledger that was designed for payment settlement and data transfer between machines and devices in the Internet of Things (IoT) ecosystem.
 The data packets that are sent through the network are called "transactions".
 Settlements or data transfers can be done with the help of these transactions. Payment settlements require the help of multiple transactions.
+
+This feature is responsible for the creation of outgoing transactions and for the interpretation of incoming transactions.
+
+
+
+# Motivation
+[...]
+
 
 # Detailed design
 
@@ -19,21 +23,24 @@ Settlements or data transfers can be done with the help of these transactions. P
 
 A transaction consists of several fields (e.g. address, value, timestamp, tag). Each field is of static length. One transaction consists of 2673 trytes:
 
-- **signature_or_message_fragment** contains the signature of the transfer or user-defined message data = 2187 trytes
-- **address** receiver (output) if value > 0, or sender (input) if value < 0 = 81 trytes
-- **value** the transferred amount in IOTA  = 27 trytes
-- **obsolete_tag** another arbitrary user-defined tag = 27 trytes
-- **timestamp** the time when the transaction was issued = 9 trytes
-- **current_index** the position of the transaction in its bundle = 9 trytes
-- **last_index** the index of the last transaction in the bundle = 9 trytes
-- **bundle hash** is the hash of the entire bundle = 81 trytes
-- **trunk hash** the hash of the first transaction referenced/approved = 81 trytes
-- **branch hash** the hash of the second transaction referenced/approved = 81 trytes
-- **nonce** is the Proof-of-Work nonce of the transaction = 27 trytes
-- **tag** arbitrary user-defined value = 27 trytes
-- **attachment_timestamp** the timestamp for when Proof-of-Work is completed = 9 trytes
-- **attachment_timestamp_lowerbound** is a slot for future use = 9 trytes
-- **attachment_timestamp_upperbound** is a slot for future use = 9 trytes
+| Name  | Description | Size|
+| ------------- | ------------- | ------------- |
+| **signature_or_message_fragment** | contains the signature of the transfer or user-defined message data | 2187 trytes |
+| **address**  | receiver (output) if value > 0, or sender (input) if value < 0  | 81 trytes |
+| **value**  | the transferred amount in IOTA  | 27 trytes |
+| **obsolete_tag**  | another arbitrary user-defined tag  | 27 trytes |
+| **timestamp**  | the time when the transaction was issued  | 9 trytes |
+| **current_index**  | the position of the transaction in its bundle  | 9 trytes |
+| **last_index**  | the index of the last transaction in the bundle  | 9 trytes |
+| **bundle hash**  | the hash of the bundle essence | 81 trytes |
+| **trunk hash**  | the hash of the first transaction referenced/approved | 81 trytes |
+| **branch hash**  | the hash of the second transaction referenced/approved | 81 trytes |
+| **nonce**  | is the Proof-of-Work nonce of the transaction | 27 trytes |
+| **tag**  | arbitrary user-defined value | 27 trytes |
+| **attachment_timestamp**  |  the timestamp for when Proof-of-Work is completed | 9 trytes |
+| **attachment_timestamp_lowerbound**  |  *not specified* | 9 trytes |
+| **attachment_timestamp_upperbound**  |  *not specified* | 9 trytes |
+
 
 A bundle is a collection of specific transactions. Bundles are required to bundle related information.
 For example, if not all data fits into one transaction, it has to be split across multiple transactions. An example would be payment settlements - they require the help of multiple transactions.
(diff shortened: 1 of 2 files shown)


text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+
+## General
+
+A transaction consists of several fields (e.g. address, value, timestamp, tag). Each field is of static length. One transaction consists of 2673 trytes:
+
+- **trunk hash** the hash of the first transaction referenced/approved = 81 trytes
+- **branch hash** the hash of the second transaction referenced/approved = 81 trytes
+- **signature_and_message_fragment** contains the signature of the transfer or user-defined message data = 2187 trytes
+- **value** the transferred amount in IOTA  = 27 trytes
+- **address** receipt (output) address if value > 0, or withdrawal (input) address if value < 0 = 81 trytes
```
tsvisabo: withdrawal => sender

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+- **bundle hash** is the hash of the entire bundle = 81 trytes
+- **current_index** the position of the transaction in its bundle = 9 trytes
+- **last_index** the total number of transactions in the bundle = 9 trytes
+
+A bundle is a collection of specific transactions. Bundles are required to bundle related information.
+For example, if not all data fits into one transaction, it has to be split across multiple transactions. An example would be payment settlements - they require the help of multiple transactions.
+A bundle then represents the collection of these transactions. It should be noted, that each transaction of a bundle will be sent separately through the network.
+
+Each transaction is identified by its hash, the transaction hash. Transactions should be final and therefore not modifiable after construction.
+The same applies to bundles. Bundles are final, and therefore not modifiable after construction.
```
tsvisabo: What about transaction's metadata? this data is not gossiped of course, but nodes need it to check/update solidity/snapshot index

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
++ Feature Name: `bee-model-crate`
++ Start Date: 2019-09-06
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#43](https://github.com/iotaledger/bee/issues/43)
+
+# Summary
+
+This feature is responsible for the creation and interpretation of transactions and bundles.
```
Alex6323: I think I'ld like the `Summary` a little bit more detailed, what is actually introduced by this feature. I know that transactions and bundles are absolutely essential to IOTA, but for consistency with other RFCs could you write this section "as if" you would introduce those concepts with this proposal. I think we can expect the reader to have "some" DLT knowledge (so this summary should still remain rather short), but IMHO we should write those RFCs also for people that are fairly new to IOTA to get as many people involved as possible. The summary/abstract will be the first thing someone interested in it will read, and most of the times the last thing as well.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+
+# Motivation
+
+IOTA is a distributed ledger that was designed for payment settlement and data transfer between machines and devices in the Internet of Things (IoT) ecosystem.
+The data packets that are sent through the network are called "transactions".
+Settlements or data transfers can be done with the help of these transactions. Payment settlements require the help of multiple transactions.
+
```
Alex6323: I think those sentences should move into the summary as they very well provide the necessary context to understand this proposal. To not repeat yourself in this section, you could pick up from the summary and elaborate on what transactions and bundles are in more detail, and how they are useful or even indispensable for the Bee framework.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+# Detailed design
+
+## General
+
+A transaction consists of several fields (e.g. address, value, timestamp, tag). Each field is of static length. One transaction consists of 2673 trytes:
```
Alex6323: You keep this somewhat inconcrete here. Is this on purpose because of future changes to the transaction design? If it isn't the case then I would just write:
An IOTA transaction on the mainnet at the time of this proposal consists of 14(?) fields, ....
Instead of 'static length' I would prefer the term 'constant length', but that's just my preference.
Additionally to mentioning trytes, I would also add the trits and bytes count, because I would have the people in mind, that are not familiar with trinary. Let's not forget that most people think in binary.

text/0000-bee-model-crate/0000-bee-model-crate.md
Alex6323: I agree with replacing `receipt`, `withdrawal` with `receiver`, `sender`.

text/0000-bee-model-crate/0000-bee-model-crate.md
Alex6323: I would suggest to not mix `transaction` and `metadata` within one struct. A transaction itself is an immutable object which should be reflected in the code. The metadata for a transaction however can change i.e. be mutated/updated, like for example from how many neighbors this transaction was received. That's why I suggest having two structs here:
`Transaction` (immutable) and `TransactionMetadata` (mutable).
Sidenote: a `Transaction` itself is also *NOT* a vertex in the Tangle. It is rather a containment relationship: `Vertex` contains reference to a `Transaction`, and to at most two other vertices in the graph.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+- **current_index** the position of the transaction in its bundle = 9 trytes
+- **last_index** the total number of transactions in the bundle = 9 trytes
+
+A bundle is a collection of specific transactions. Bundles are required to bundle related information.
+For example, if not all data fits into one transaction, it has to be split across multiple transactions. An example would be payment settlements - they require the help of multiple transactions.
+A bundle then represents the collection of these transactions. It should be noted, that each transaction of a bundle will be sent separately through the network.
+
+Each transaction is identified by its hash, the transaction hash. Transactions should be final and therefore not modifiable after construction.
+The same applies to bundles. Bundles are final, and therefore not modifiable after construction.
+Having bundles as well as transactions final helps achieve correctness among the data objects.
```
Alex6323: I don't understand this sentence. How is 'correctness among data objects' defined? I have an idea of what you mean here, but can you make this more clear so someone from the outside can follow you here?

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+
+Transactions can be built from Transaction Builders.
+Bundles can be built from Bundle Builders. Both builder objects can be manipulated as desired.
+It should be noted, Bundle Builders don't process Transactions (as transactions are final), instead, they process Transaction Builders.
```
Alex6323: what do you mean by 'they process transaction builder'? Aren't transaction builders used to simplify building transactions? Why would they needed to be processed. Maybe jus a wording thing, but I have different associations with the verb 'to process'.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+    trunk: String,
+    branch: String,
+    nonce: String,
+    tag: String,
+    attachment_timestamp: i64,
+    attachment_timestamp_lower_bound: i64,
+    attachment_timestamp_upper_bound: i64,
+
+}
+```
```
Alex6323, line 132: I suggest a different approach here, because I see two main issues:
1. the transaction fields are (by design) of constant size known upfront. We should right of the bat make use of this knowledge, and store those fields in fixed sized arrays (if we go that route which I wouldn't suggest as it involves lots of allocations -> see proposal below)
2. this proposal of a transaction seems to be used for eagerly deserializing the transaction data from the received gossip (T5B1 or T9B2) bytes, which probably involves unnecessary calling of conversion functions in the hot-path, e.g. converting timestamps (i64). I haven't implemented a lot of IOTA nodes, but wouldn't it make more sense to only deserialize transaction fields lazily, that is on-demand? This might be more painful with T5B1 encoding, but with T9B2 each field has an integer byte size as well, which makes partial deserializing a breeze. That's why maybe it would be better to model `Transaction like so:
```Rust
struct Transaction {
    raw: Trits<T9B2>,
}
impl Transaction {
    ...
    fn address(&self) -> String {/*deserialize at specific location in self.raw*/}
    fn value(&self) -> i64 {/*same*/.}
    ...
}
``` 
It is also possible to choose a route in the middle of both extremes, where fields, that are always required, are deserialized eagerly. Just add fields for those to the `Transaction` struct (e.g. trunk, branch). If those T9B2 encoded trits are also used for transport (e.g. among Bee nodes) then in-memory copy operations could be reduced by a lot.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+
+As mentioned, Transactions are final. Transaction fields therefore should be only accessible by getter functions.
+Besides that, Transactions can be only built by Transaction Builders, or by the bytes of a received transaction.
```
Alex6323: Hopefully I don't come across as too nit-picky, but that sentence seems a bit vague to me. Suggestion:
"transactions can only be created from a transaction builder, or from a constructor function which expects encoded bytes as received e.g. from a network socket."

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+
+```rust
+impl Transaction {
+
+    pub fn from_transaction_builder(transaction_builder: &TransactionBuilder) -> Result<Transaction, TransactionBuilderValidationError> {
```
Alex6323: Suggestion: Do not let the `Transaction` struct know about a `TransactionBuilder`. There are two reasons:
1. Semantically: It's as if a car knows about its fabrication facility, but ideally a car shouldn't care about where it's produced. In the hierarchy of objects the `Transaction` is clearly subordinate to a `TransactionBuilder` which handles/operates on many transactions.
2. Breaking encapsulation/Dependency hell: IMHO one should only very carefully make types known to other types, and prevent it at all costs if possible. Refactoring code in one place should ideally not require changes in other parts.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+## Exposed Interface
+
+### Transaction
+
+ A transaction consists of following fields:
+
+```rust
+struct Transaction {
+
+    transaction_hash: String,
```
Alex6323: the hash of a transaction can be considered metadata of a `Transaction` and should not be part of the `Transaction` struct which itself is an immutable object.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+impl Transaction {
+
+    pub fn from_transaction_builder(transaction_builder: &TransactionBuilder) -> Result<Transaction, TransactionBuilderValidationError> {
+                
+        match TransactionBuilderValidator::validate(&transaction_builder) {
+            Ok(()) => { ; }
+            Err(e) => { return Err(e) }
+        }
+
+        let (transaction_hash, nonce, timestamp) = Pow::compute(&transaction_builder);
```
Alex6323: IMHO this should be designed in a way that `Pow::compute(...)` returns a tuple, in this case a pair `(transaction, metadata)` where metadata contains the `transaction_hash` and the `timestamp`.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,279 @@
...
+            last_index: transaction_builder.last_index.clone(),
+            bundle_hash: transaction_builder.bundle_hash.clone(),
+            trunk: transaction_builder.trunk.clone(),
+            branch: transaction_builder.branch.clone(),
+            nonce,
+            tag: transaction_builder.tag.clone(),
+            attachment_timestamp: transaction_builder.attachment_timestamp.clone(),
+            attachment_timestamp_lower_bound: transaction_builder.attachment_timestamp_lower_bound.clone(),
+            attachment_timestamp_upper_bound: transaction_builder.attachment_timestamp_upper_bound.clone(),
+        })
```
Alex6323: I think this way of creating the transaction is dangerous. I think a TransactionBuilder should be destroyed having a method `build(self) -> Transaction` moving the internal state (the transaction to be built) out. I seems like what you want to have is a `TransactionFactory`, that has some template and creates clones off of it. But since transactions are completely unique instances where nothing can be really shared across them, the factory pattern doesn't seem to be a good choice here.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: Agree @Alex6323, we should separate `metadata` from `transaction`. We don't want to create a [god object](https://en.wikipedia.org/wiki/God_object).

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: Thanks for the comment. Agree, moving ownership and destroy it afterwards seems to be the better choice.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: Timestamp should be part of the transaction not metadata. `(transaction, metadata)` would imply, you would like to do the creation of a `Transaction` in `Pow`.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: What we can do is moving initialization of the `Transaction` struct directly in the `build(self)` of the `TransactionBuilder`.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: I'm talking in the sense of software quality, as defined here:
"Correctness: The degree to which a system is free from [defects] in its specification, design, and implementation." [https://accu.org/index.php/journals/1585](https://accu.org/index.php/journals/1585)
Objects that contain properties/features that shouldn't be there. It's not meant in the sense of functional correctness regarding input-output behavior of an algorithm.
Will paraphrase it.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: > Why would they needed to be processed.
Solved in [https://github.com/iotaledger/bee-rfcs/pull/3#discussion_r329781471](https://github.com/iotaledger/bee-rfcs/pull/3#discussion_r329781471)

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: Agree you fully, let's wait for the others what they think.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: Btw, if we do the creation of `Transaction` in Pow, we need to declare the Transaction fields as public. Else we end in something like this [https://doc.rust-lang.org/error-index.html#E0451](https://doc.rust-lang.org/error-index.html#E0451)
@Alex6323 moved the actual transaction struct initialization into Pow

text/0000-bee-model-crate/0000-bee-model-crate.md
SuperFluffy: I prefer language such as:

> This feature proposes the `Transaction` type to translate an incoming IOTA IRI transaction into a Rust data model.

Also, transactions and bundles seem to be two different concerns, operating at different levels of abstractions. I propose opening a separate RFC for transactions.

text/0000-bee-model-crate/0000-bee-model-crate.md
SuperFluffy: I agree with @Alex6323 here.

text/0000-bee-model-crate/0000-bee-model-crate.md
SuperFluffy: I also agree with @Alex6323 on this one. Ideally, we would be able to refer to some kind of document, spec, or just an SHA hash (of a github commit) where this message format is introduced. In the absence of that, one could just say that one wants to implement the message protocol as used by IRI (and cIRI?) in their most recent releases (reference the release, the git tag, and/or an SHA hash here).


Regarding the ternary encoding: I think it's fine mentioning that the incoming transaction is encoded in ternary, but for all intents and purposes it is not material what the specific encoding is. So, the fields of the `Transaction` `struct` should either be proper parsed Rust types, or they should be left in whichever encoding they are in when coming over the wire, but just extracted from the general message.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+- **timestamp** the time when the transaction was issued = 9 trytes
+- **attachment_timestamp** the timestamp for when Proof-of-Work is completed = 9 trytes
+- **attachment_timestamp_lowerbound** is a slot for future use = 9 trytes
+- **attachment_timestamp_upperbound** is a slot for future use = 9 trytes
+- **tag** arbitrary user-defined value = 27 trytes
+- **obsolete_tag** another arbitrary user-defined tag = 27 trytes
+- **nonce** is the Proof-of-Work nonce of the transaction = 27 trytes
+- **bundle hash** is the hash of the entire bundle = 81 trytes
+- **current_index** the position of the transaction in its bundle = 9 trytes
+- **last_index** the total number of transactions in the bundle = 9 trytes
```
SuperFluffy: I am wondering if we can make a nice ASCII art out of this.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+
+A bundle is a collection of specific transactions. Bundles are required to bundle related information.
```
SuperFluffy: I think it's fine to mention bundles, but they should probably be in their own RFC to keep the focus on the transaction itself.

text/0000-bee-model-crate/0000-bee-model-crate.md
SuperFluffy: If we indeed know the size of each field a priori, then we should model it accordingly. In addition, I think we should probably introduce a type for each of these fields in case they have to uphold invariants themselves.

Also, an array of `T`s with `m` elements, `[T; m]`, is allocated on the stack in Rust. If we want to have vec semantics, there is also the `smallvec` crate written by the Servo project.

As regards the `T9B2` encoding, we have discussed this yesterday (so after Alex left his review comment here) that we will for now stay within the already existing binary-coded trinary representations.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+    pub fn attachment_timestamp_lower_bound(&self) -> &i64 {
+        &self.attachment_timestamp_lower_bound
+    }
+
+    pub fn attachment_timestamp_upper_bound(&self) -> &i64 {
+        &self.attachment_timestamp_upper_bound
+    }
+
+}
+```
```
SuperFluffy, line 163: I think you can safely remove most of the getter functions above to demonstrate what you mean. :) Let's mention that the fields should be left private and immutable.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+
+The Transcation_Builder struct contains public accessible fields.
```
SuperFluffy: `Transcation_Builder -> TransactionBuilder`

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+    pub current_index: usize,
+    pub last_index: usize,
+    pub bundle_hash: String,
+    pub trunk: String,
+    pub branch: String,
+    pub tag: String,
+    pub attachment_timestamp: i64,
+    pub attachment_timestamp_lower_bound: i64,
+    pub attachment_timestamp_upper_bound: i64,
+}
```
SuperFluffy: For a transaction builder, the proper semantics are to keep the fields as `Option<T>`. Also, I would also leave the fields `non-pub` and only allow setting them through setter methods.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+
+The Pow object takes the Transaction Builder as argument and returns the built transaction together with its metadata.
+Every transaction contains a mutable metadata which consists of following attributes:
+```rust
+pub struct TransactionMetadata {
+    pub transaction_hash: String,
+    pub is_solid: bool,
+    pub usize: snapshotIndex
+}
+```
```
SuperFluffy, line 175: As before, I would leave this `non-pub`. Since we are probably interested in upholding invariants, we shouldn't give out direct access to the fields.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+
+Even though Proof-Of-Work should be handled in its own RFC, this example shows how it could interact with the TransactionBuilder and how Transactions could be built.
+This example assumes that the actual Transaction building takes place here.
+
+
+```rust
+pub struct Pow;
+
+impl Pow {
```
SuperFluffy: Let's move all this into its own section and call it “how this will be used” or something like that.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+        )
+
+    }
+
+}
+
+```
+
+
+### Bundle
```
SuperFluffy: This should go into its own RFC.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+}
+
+```
+
+
+# Drawbacks
+
+Without any bee-model crate, nodes can not exchange transactions. Therefore this crate seems necessary.
```
SuperFluffy: Drawbacks should ideally also discuss the drawbacks of this design. :)

I am thinking of the `String` types, for example, which could also be done as their own types.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+
+# Rationale and alternatives
```
SuperFluffy, line 246: An additional rationale is definitely that this kind of interface is relatively minimal and easily extended upon in a future iteration.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+
+- The distinction between Transaction and Transaction Builder as well as Bundle and Bundle Builder makes the code cleaner 
+and helps achieve correctness among the data objects. Properties are clearly assigned to specific data objects and not mixed up.
+
+- The proposed crate interface is intuitive. Completely different alternatives did not naturally come to mind.
```
SuperFluffy: If it's intuitive is for users to decide. I agree however that it is a very natural proposition to basically treat the incoming transaction as some encoded binary blob.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+
+- The validation logic could be done in a separate validator class, which then will be called in the build() function.
+
+# Unresolved questions
+
+- How is Proof-Of-Work acceleration handled?
```
SuperFluffy: This should probably be further explained. Is `Proof-of-Work acceleration` really material to this proposal? If not, I would leave it out.

Important unresolved questions:

+ How to handle deserialization of the incoming, encoded transaction?

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
++ Feature Name: `bee-model-crate`
```
SuperFluffy: We should rename it to something like `iri-transaction-model` or `mainnet-transaction-model` or `v1-transaction-model`.

SuperFluffy (changes requested):
I like this RFC because I understand its purpose, and because I feel like it fulfills (or is very close to fulfilling) the requirement of taking care of a specific, cleanly separated concern.

The proposal should definitely separate the `Transaction` and `Bundle` types into their respective RFCs, because these seem to be sufficiently different.

The RFC can mention that the `Transaction` type could live in its own crate.

text/0000-bee-model-crate/0000-bee-model-crate.md
Alex6323: @rufsam yea, your're right. I don't know why I wrote the timestamp would be metadata. I mean it is kind of, right?, but as it's also part of the transaction itself, of course it would be part of it.

text/0000-bee-model-crate/0000-bee-model-crate.md
thibault-martinez:
```suggestion
+ Feature name: `bee-model-crate`
```
The template changed since you opened.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
++ Start Date: 2019-09-06
```
thibault-martinez:
```suggestion
+ Start date: 2019-09-06
```
The template changed since you opened.

text/0000-bee-model-crate/0000-bee-model-crate.md
thibault-martinez: I'd vote for versioning since `mainnet-transaction-model` is not time-proof and `iri-transaction-model` is not really the case.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
```
thibault-martinez:
```suggestion
+ RFC PR: [iotaledger/bee-rfcs#3](https://github.com/iotaledger/bee-rfcs/pull/3)
```

text/0000-bee-model-crate/0000-bee-model-crate.md
thibault-martinez: What about a table Name | Description |  Size ?

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+
+# Detailed design
+
+## General
+
+A transaction consists of several fields (e.g. address, value, timestamp, tag). Each field is of static length. One transaction consists of 2673 trytes:
+
+- **trunk hash** the hash of the first transaction referenced/approved = 81 trytes
+- **branch hash** the hash of the second transaction referenced/approved = 81 trytes
+- **signature_and_message_fragment** contains the signature of the transfer or user-defined message data = 2187 trytes
```
thibault-martinez: `signature_or_message_fragment` ?

text/0000-bee-model-crate/0000-bee-model-crate.md
thibault-martinez: I know I'm being picky here but can you list  them in their actual order ? I really think it helps a lot.

thibault-martinez: That is not true. The total number of transactions in the bundle is `last_index + 1`.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+## Exposed Interface
+
+### Transaction
+
+ A transaction consists of following fields:
+
+```rust
+struct Transaction {
+
+    signature_fragments: String,
```
thibault-martinez: I'd like the name to reflect that it doesn't necessarily contain a signature. `signature_or_message` ?

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+    bundle_hash: String,
+    trunk: String,
+    branch: String,
+    nonce: String,
+    tag: String,
+    attachment_timestamp: i64,
+    attachment_timestamp_lower_bound: i64,
+    attachment_timestamp_upper_bound: i64,
+
+}
```
thibault-martinez: Still being picky, can you order the fields ?

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+    current_index: usize,
+    last_index: usize,
+    bundle_hash: String,
+    trunk: String,
+    branch: String,
+    nonce: String,
+    tag: String,
+    attachment_timestamp: i64,
+    attachment_timestamp_lower_bound: i64,
+    attachment_timestamp_upper_bound: i64,
```
thibault-martinez: Timestamps are unsigned

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,334 @@
...
+```
+
+The Pow object takes the Transaction Builder as argument and returns the built transaction together with its metadata.
+Every transaction contains a mutable metadata which consists of following attributes:
+```rust
+pub struct TransactionMetadata {
+    pub transaction_hash: String,
+    pub is_solid: bool,
+    pub snapshot_index: usize
+}
```
thibault-martinez, line 174: I think `TransactionMetadata` will be the topic of another RFC so I wouldn't necessarily put the structure here. Or at least not `is_solid` and `snapshot_index`.

text/0000-bee-model-crate/0000-bee-model-crate.md
thibault-martinez: Yeah not sure we  should mention PoW at all here.

thibault-martinez (changes requested):

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: How do you want to order them?

thibault-martinez: Order in which the fields appear in a transaction.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: How do you want to order them?

thibault-martinez: Order in which the fields appear in a transaction.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: How do you want to version it?

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: There was an idea to build the Transaction directly in the compute of PoW. If we go with it, IMO we need to show what role PoW has and how the Transaction gets finally built.

samuel-rufi: @SuperFluffy Agree

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: I'm not sure if we need an additional timestamp in the metadata. I think we are fine with having it only in the transaction.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: > For a transaction builder, the proper semantics are to keep the fields as Option<T>

Great idea

> Also, I would also leave the fields non-pub and only allow setting them through setter methods.

Where would you put the validation of the set values? Directly in the setter, or as it currently is in `build()`? Currently the validation happens one time, else it could be done in each setter directly. Not sure what suites best here. If we decide in each setter, we need to keep in mind that every setter could throw an error and how handy it is for the programmer.

samuel-rufi: Regarding `non-pub` there is the idea to build the actual Transaction (creating the struct) in Pow::compute(). Currently we have the Transaction fields `non-pub`. It seems like this ends in something like this https://doc.rust-lang.org/error-index.html#E0451

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,290 @@
...
+### Transaction
+
+ A transaction consists of following fields:
+
+```rust
+struct Transaction {
+
+    signature_fragments: String,
+    address: String,
+    value: u64,
```
thibault-martinez:
```suggestion
    value: i64,
```

thibault-martinez (changes requested):

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
++ Feature name: `bee-transaction-crate`
```
thibault-martinez: Not a full crate

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
...
+| ------------- | ------------- | ------------- |
+| **signature_or_message_fragment** | contains the signature of the transfer or user-defined message data | 2187 trytes |
+| **address**  | receiver (output) if value > 0, or sender (input) if value < 0  | 81 trytes |
+| **value**  | the transferred amount in IOTA  | 27 trytes |
+| **obsolete_tag**  | another arbitrary user-defined tag  | 27 trytes |
+| **timestamp**  | the time when the transaction was issued  | 9 trytes |
+| **current_index**  | the position of the transaction in its bundle  | 9 trytes |
+| **last_index**  | the index of the last transaction in the bundle  | 9 trytes |
+| **bundle hash**  | the hash of the bundle essence | 81 trytes |
+| **trunk hash**  | the hash of the first transaction referenced/approved | 81 trytes |
```
thibault-martinez:
```suggestion
| **trunk**  | the hash of the first transaction referenced/approved | 81 trytes |
```

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
...
+| **branch hash**  | the hash of the second transaction referenced/approved | 81 trytes |
```
thibault-martinez:
```suggestion
| **branch**  | the hash of the second transaction referenced/approved | 81 trytes |
```

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
...
+| **nonce**  | is the Proof-of-Work nonce of the transaction | 27 trytes |
```
thibault-martinez: Sorry to bother again with this but `nonce` is the last field :p

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
...
+| **tag**  | arbitrary user-defined value | 27 trytes |
+| **attachment_timestamp**  |  the timestamp for when Proof-of-Work is completed | 9 trytes |
+| **attachment_timestamp_lowerbound**  |  *not specified* | 9 trytes |
+| **attachment_timestamp_upperbound**  |  *not specified* | 9 trytes |
+
+As reflected in the table, each field is of constant length. Totally, one transaction consists of **2673 trytes**.
+Transactions should not be **modifiable** after construction. Transaction fields therefore only should be accessible by getter functions.
```
thibault-martinez: Can you expand a little bit on why it should not be modifiable ?  We can't just say it should not.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
...
+    address: String,
+    value: i64,
+    obsolete_tag: String,
+    timestamp: u64,
+    current_index: usize,
+    last_index: usize,
+    bundle_hash: String,
+    trunk: String,
+    branch: String,
+    nonce: String,
```
thibault-martinez: `nonce` is the last field.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
...
+| Name  | Description | Size |
+| ------------- | ------------- | ------------- |
+| **signature_or_message_fragment** | contains the signature of the transfer or user-defined message data | 2187 trytes |
+| **address**  | receiver (output) if value > 0, or sender (input) if value < 0  | 81 trytes |
+| **value**  | the transferred amount in IOTA  | 27 trytes |
+| **obsolete_tag**  | another arbitrary user-defined tag  | 27 trytes |
+| **timestamp**  | the time when the transaction was issued  | 9 trytes |
+| **current_index**  | the position of the transaction in its bundle  | 9 trytes |
+| **last_index**  | the index of the last transaction in the bundle  | 9 trytes |
+| **bundle hash**  | the hash of the bundle essence | 81 trytes |
```
thibault-martinez:
```suggestion
| **bundle**  | the hash of the bundle essence | 81 trytes |
```

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
...
+struct Transaction {
+
+    signature_fragments: String,
+    address: String,
+    value: i64,
+    obsolete_tag: String,
+    timestamp: u64,
+    current_index: usize,
+    last_index: usize,
+    bundle_hash: String,
```
thibault-martinez: I'd just call that `bundle`

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,264 @@
...
+
+}
+```
+
+The **from_bytes(bytes: &Vec)** is striking here. It serves as constructor and allows to build transactions from received bytes, e.g. from the network socket.
+It's one way to build transactions. Another way to build transactions would be via **TransactionBuilder**.
+
+### TransactionBuilder
+
+The TransactionBuilder offers a **simple and convenient** way to built transactions. It's defined as follows:
```
thibault-martinez:
```suggestion
The TransactionBuilder offers a **simple and convenient** way to build transactions. It's defined as follows:
```

thibault-martinez (changes requested):

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: Then we probably also should call `trunk hash` `trunk` and `branch hash` `branch`

thibault-martinez: They are already called `trunk` and `branch`

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: lets call it module?

lunfardo314:
`Transaction` and related types better be abstract traits since the beginning.
I refer to requirements of Q-nodes to Bee: https://github.com/iotaledger/bee-rfcs/issues/6#issue-494583275 . 
The assemblies would need multiple types of transactions in the same bundle and some new constrains on the ledger, other than usual constraints on token accounts. For example, there's a need to support unforkable `chains`, signed by the assembly.

text/0000-bee-model-crate/0000-bee-model-crate.md
wusyong: How about this?
```suggestion
+ Feature name: `transaction-module`
```

text/0000-bee-model-crate/0000-bee-model-crate.md
wusyong: Is it possible to even not contain this field if the instance doesn't have `signature_or_message`? It would be sth like this:
```suggestion
    signature_or_message: Option<String>,
```

text/0000-bee-model-crate/0000-bee-model-crate.md
wusyong: I'm also agree this. The return value can be what @SuperFluffy said, a fixed size array, or I think maybe it's also possible to be a reference slice.

text/0000-bee-model-crate/0000-bee-model-crate.md
wusyong: Let's use visibility we introduced yesterday then. I think `pub(crate)` should be fine.

text/0000-bee-model-crate/0000-bee-model-crate.md
thibault-martinez: `Option` still allocates memory for the field so it's not really helpful

samuel-rufi: @SuperFluffy what are your thoughts regarding `Option`. Semantically, it would be more correct, right?

thibault-martinez: I'm not sure it makes more sense "semantically", an empty sig_or_msg doesn't mean there is no sig_or_msg. There is always a sig_or_msg :)

samuel-rufi: Agree, so maybe Option isn't necessary. And if we use static types, we could initialize it with an empty (9's) `SignatureOrMessageFragment` and could be sure the internal state of the builder object is always valid.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -146,7 +170,7 @@ Each transaction contains a mutable metadata which is defined as following:
 
 ```rust
 pub struct TransactionMetadata {
-    pub transaction_hash: String,
+    pub transaction_hash: TransactinHash,
```
thibault-martinez: Typo

thibault-martinez (changes requested):

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: To add the bytes count, we need to sort the encoding topic. I will add trits count. "static length" is replaced by now, also the transaction is defined in more detail. Also the field types are replaced. Will look the for the SHA hash of the tx layout were it was changed last time.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -299,8 +296,6 @@ Without any bee-model crate, nodes can not exchange transactions. Therefore this
 
 - Where should the actual Transaction struct be initialized/returned? Current ideas are to put it in the build() of the TransactionBuilder or as it currently is in the compute() of Pow.
 
-- Where should Transaction fields get validated? One-time validation once build() is called by the separate *TransactionBuilderValidator* as it currently is, or use **setters** in the TransactionBuilder and validate in each setter separately?
-
-- Should we use setters for the TransactionBuilder?
+- Should we introduce a fluent API to the TransactionBuilder?
 
-- Introducing a fluent API to the TransactionBuilder?
+- Should we use Option<T> for the fields in TransactionBuilder?
```
wusyong: Some fields would need this I guess. Others like timestamp, index etc are mandatory.

SuperFluffy: The `build` method is responsible for ensuring that mandatory fields are set, returning `Result::Err` if not. The type of the field is `Option` if there is no natural (read: meaningful) way of providing a default.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,277 @@
...
+
+A transaction is a sequence of 15 fields of constant length. A transaction has
+a total length of 8019 trits. The fields' name, description, and size (in
+trits) is summarized in the table below, in their order of appearance in
+a transaction.
+
+| Name                              | Description                                            | Size (in trits) |
+| ---                               | ---                                                    | ---             |
+| `signature_or_message_fragment`   | contains the signature of the transfer                 |                 |
+|                                   | or user-defined message data                           | 6561            |
```
wusyong, line 68: I think it's no need to be separate line. Besides, this would become totally different row. Normal markdown display would just adjust them.

```suggestion
| `signature_or_message_fragment` | contains the signature of the transfer or user-defined message data | 6561 |
```

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,277 @@
...
+| `address`                         | receiver (output) if value > 0,
+|                                   | or sender (input) if value < 0                         | 243             |
```
wusyong, line 70: ditto

SuperFluffy:
Yeah, I wanted to stay within the number of max chars per line we discussed.

I thought GitHub markdown supports multiline table cells, but I either
chose the wrong table format or GitHub doesn't support these kinds of raw
tables as pandoc does.

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,277 @@
...
+The `from_slice` and `from_reader` methods will use a `TransactionDraft` internally,
+but these are implementation details.
+
+The `TransactionError` is not fully fleshed out yet. For the time being, it
+definitely contains `io::Error`:
+
+```rust
+pub enum TransactionError {
+    Io(io::Error),
+}
```
wusyong, line 174: I think we should also start considering which error handling library to use. I saw there's another new one called [thiserror](https://github.com/dtolnay/thiserror). What do you think?

text/0000-bee-model-crate/0000-bee-model-crate.md
```diff
@@ -0,0 +1,277 @@
...
+    pub fn from_slice(stream: &[u8]) -> Result<Self, TransactionDraftError> {
+        unimplemented!()
+    }
+
+    /// Verify that all fieleds are set and build the `Transaction`.
+    pub fn build(self) -> Result<Transaction, TransactionDraftError> {
+        unimplemented!()
+    }
+
+    pub fn address(&mut self, address: Address) -> &mut Self {
```
wusyong, line 217: How about give a proper setter method prefix like `set_address`?

text/0000-bee-model-crate/0000-bee-model-crate.md
wusyong: @SuperFluffy I think we could ignore max chars limit during table format. We should make representation correct at least.

text/0000-bee-model-crate/0000-bee-model-crate.md
samuel-rufi: Yes, as RFC #344 notes it:
[https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis)

SuperFluffy:
Incorporated into #20.

While discussing the design of the `Transaction`, we realized that a new transaction is not supposed to be create directly. Instead, when constructing a new message, the `Bundle` is the fundamental object one wants to crate, which in turn splits up its message into several `Transaction`s. A `Transaction` thus becomes a very simple type. More complicated logic should thus be handled by the `Bundle`.

PR #4, opened 2019-09-08
RFC: implement type tryte in trinary

wusyong:
[Rendered](https://github.com/wusyong/bee-rfcs/blob/trytes/text/0000-trinary-trytes/0000-trinary-trytes.md)


text/0000-trinary-trytes.md
```diff
@@ -0,0 +1,202 @@
...
+
+Note that `vec` is not public, we provides plenty of methods and even implment
+traits needed in most scenarios. Many of them are similar to `String`'s but have
+different signatures mainly for checking if inputs are actaully valid. So
+essentailly, these would be literally same methods but with `Result` returned.
+The methods and traits are going to br implemented are roughly list below:
+
+```rust
+impl Trytes {
+    #[inline]
```
zesterer: `#[inline]` is generally not recommended since it can hurt compilation times. See https://internals.rust-lang.org/t/when-should-i-use-inline/598/3

text/0000-trinary-trytes.md
```diff
@@ -0,0 +1,202 @@
...
+    pub fn clear(&mut self) {
+    }
+
+    unsafe fn insert_bytes(&mut self, idx: usize, bytes: &[u8]) {
+    }
+
+    fn all_tryte_alphabete<'a, I>(vals: I) -> Result<()>
+    where
+        I: Iterator<Item = &'a u8>,
```
zesterer: It would probably be better if `I: Iterator<Item=u8>` (simply because this conversion would be occurring inside this method regardless). Going from an iterator that produces references to one that produces `u8`s by value is just a matter of calling `.copied()` on the iterator.

text/0000-trinary-trytes.md
```diff
@@ -0,0 +1,180 @@
...
+essentailly, these would be literally same methods but with `Result` returned.
+The methods and traits are going to br implemented are roughly list below:
+
+```rust
+impl Trytes {
+    pub fn new() -> Trytes {
+        Trytes { vec: Vec::new() }
+    }
+
+    pub fn with_capacity(capacity: usize) -> Trytes {
```
vuapo-eth: What's the rationale for making it variably-sized? Are there use cases where the sizes would change? Have we considered fixed sizes? I see at least one potential advantage of fixed sizes:
- Typing: Every size has a type. This way we can ensure at compile time that an argument has the required size. For example `fn find_transactions_by_address(address : Trit243) -> Vec<Transaction>` can only be called with a 243-trit long argument. This reduces bugs and increases performance as we do not require a separate runtime check.

text/0000-trinary-trytes.md
```diff
@@ -0,0 +1,180 @@
...
+
+# Detailed design
+
+## Trytes
+
+Type `Trytes` is a vector of valus `Vec<u8>` and moreover a smart pointer.
+The structure simply looks like this:
+
+```rust
+pub type Tryte = u8;
```
vuapo-eth: Using a byte per tryte is certainly the most intuitive solution. @Alex6323 and I had an alternative idea that would increase speed and reduce memory: encode 3 trytes with 2 bytes:
* this reduces memory requirements by 33%
* we can compare vectors faster as we compare less bytes
* we do not need to compress and decompress trytes when receiving/sending transaction to neighbors but can simply forward the internal representation

vuapo-eth (commented):
Added two thoughts on alternative designs. Should be considered in comparison to the current proposal.

text/0000-trinary-trytes.md
wusyong: This is proposal for trytes only as trits has more encoding decision need to consider. Capacity is not length, vector is allocated according to this. I think it's handy for trytes(not trits) be able to reallocate.

text/0000-trinary-trytes.md
wusyong: You are talking trits encoding which we will definitely have various encoding implementations like TE27 used in troika. I would like to put encoding and operation logic to trits only. So trytes can be easy to understand and serves as tryte alphabet. We still need to convert between different encodings afterwards though.

text/0000-trinary-trytes.md
```diff
@@ -0,0 +1,180 @@
...
+
+pub struct Trytes {
+    vec: Vec<Tryte>,
```
SuperFluffy: Do you plan to add more fields to this struct? Otherwise I would probably use the [`newtype`](https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html) pattern, i.e. a tuple struct:

```rust
struct Tytes(Vec<Tryte>)
```

text/0000-trinary-trytes.md
SuperFluffy: So my understanding is that the `Trytes` type is for data of any length, right?

I guess we can represent transactions as [`smallvec`](https://crates.io/search?q=smallvec), @mikrohash.

zesterer:
@SuperFluffy and I were talking about the possibility of specifying different encodings with minimal hassle. Borrowing similar ideas from `RawVec` (part of Rust's allocator API), we imagined a `RawTrytes` trait that would specify a minimal API that tryte encodings must implement. We'd then keep the `Trytes` struct (now `Trytes<T: RawTrytes>`) as a high-level wrapper with an ergonomic API around whatever encoding was desired.

Here's a rough example of how that design would look, with an example encoding (1 byte = 1 tryte).

```rust
trait RawTrytes {
    fn with_capacity(n: usize) -> Self;
    fn double_in_place(&mut self) -> bool;
    fn reserve(&mut self, used: usize, needed: usize);
    ...
}

struct RawTrytes1B1T {
    trytes: Vec<u8>,
}

impl RawTrytes for RawTrytes1B1T {
    ...
}

// High-level wrapper interface

struct Trytes<T: RawTrytes = RawTrytes1B1T> {
    raw: T,
}

impl<T: RawTrytes> Trytes<T> {
    pub fn new() -> Self { ... }
    
    pub fn with_capacity(n: usize) -> Self { ... }
    
    pub fn push(&mut self, x: Tryte) { ... }
    
    pub fn pop(&mut self) -> Option<Tryte> { ... }
}
```

If this seems a little overkill, then there's no reason to implement it yet. The nice thing is that the default parameter means that this can be implemented later without a breaking API change.

wusyong:
@zesterer Thanks! I think this might be the solution we are looking for `Trits` implementation. As for this RFC, I really couldn't think of any case than one tryte per byte. Even if it is 9 trits per 2 bytes and we want to represent. It will end up with string literal which is one per byte. I think we could stick with current structure for now, unless there's a critical case.

text/0000-trinary-trytes.md
wusyong: Thanks, I will implement like this instead.

text/0000-trinary-trytes.md
vuapo-eth: I'm not arguing about the implementation but the interface. I think we can increase correctness and performance if the trinary type for addresses which 81 trytes / 243 trits long, is different from the type for transaction tags with a length of 27 trytes / 81 trits.

text/0000-trinary-trytes.md
vuapo-eth: What about `Trytes::from_human_readable("SOME9TRYTE9SEQUENCE");`? Internal representation could still be compressed to 9 trits per 2 bytes to reduce memory requirements and increase performance (as no compression/decompression is required for transmission to neighbors).

wusyong: This implementation is for human readable strings. All encoding technique I encourage stay in any self-defined trit data structure (that say you have a trit type called `9TritsPer2Bytes`). I think the internal of bee would no need for `Trytes`. Unless you want to repressent or debug, in this case you can define a method of your `9TritsPer2Bytes` to convert to `Trytes`.

vuapo-eth: Ah, perfect. But do we actually need a type just for human readable strings? Wouldn't `CompactTrinary::to_human_readable(&self) -> String` suffice, where `CompactTrinary` is the type that internally stores the 9 trit / 2 byte encoding?

wusyong: Yes, it's defined as `String` in `iota-lib-rs`. But since we are going to re-implement with TDD, I'm thinking to define a more secure one. The methods I proposed have almost same signatures with string but many are return with `Result` type to make sure it's valid input. But indeed some might be overkill, I'm going to reduce some with zesterer's advice.

text/0000-trinary-trytes.md
thibault-martinez: I very much agree with @mikrohash on this. In what situation do we need trytes that are not of usual sizes (9, 27, 81, 2187 or 2673) ? Having sizes checked at compile-time seems to be a very good idea to me.

thibault-martinez: We could even typedef fields to these sized trytes: Tag <-> Trytes27, Address <-> Trytes81 . etc...

text/0000-trinary-trytes/0000-trinary-trytes.md
```diff
@@ -0,0 +1,126 @@
++ Feature Name: `trinary-trytes`
++ Start Date: 2019-09-09
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/4)
```
thibault-martinez:
```suggestion
+ RFC PR: [iotaledger/bee-rfcs#4](https://github.com/iotaledger/bee-rfcs/pull/4)
```

text/0000-trinary-trytes/0000-trinary-trytes.md
```diff
@@ -0,0 +1,126 @@
...
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/50)
```
thibault-martinez:
```suggestion
+ Bee issue: [iotaledger/bee#50](https://github.com/iotaledger/bee/issues/50)
```

thibault-martinez (changes requested):

text/0000-trinary-trytes/0000-trinary-trytes.md
```diff
@@ -0,0 +1,126 @@
...
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0.as_ref()
+    }
+
+    ...
+
+    pub fn encode(&mut self, plain: &str) {
+        for byte in plain.bytes() {
+            let first = byte % 27;
```
zesterer: Should this code accept out-of-range characters so readily? I think this method should return a `Result<(), Error>` - if nothing else, as a sanity check.

text/0000-trinary-trytes/0000-trinary-trytes.md
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    pub fn as_bytes(&self) -> &[u8] {
+        &self.0.as_ref()
+    }
+
+    ...
+
+    pub fn encode(&mut self, plain: &str) {
+        for byte in plain.bytes() {
```
zesterer: I think iterating upon `plain.chars()` is better - not least because the current implementation accepts unicode and would decode some characters as multiple trytes.

text/0000-trinary-trytes/0000-trinary-trytes.md
```diff
@@ -0,0 +1,126 @@
...
+            let first = byte % 27;
+            let second = (byte - first) / 27;
+            self.0.put(TRYTE_ALPHABET[first as usize]);
+            self.0.put(TRYTE_ALPHABET[second as usize]);
+        }
+    }
+
+    pub fn decode(&self) -> String {
```
zesterer, line 79: Would it be possible to explain a little more about what this method is doing? It appears to decode arbitrary trytes into a representation that includes ASCII control characters, extended ASCII, etc. I would think that such a representation isn't well-suited to `String`.

wusyong: I will also answer the question about encode above. This is originally going to support ASCII character only, but I thought it can do more like encode any byte. Indeed I didn't consider that much about safety. It can switch back to ascii only, or we refactor a bit to encode/decode any byte and let user decide how they convert these bytes.

samuel-rufi: It would be great to convert bytes to trytes without going the unsafe String::from_utf8_unchecked() route.

text/0000-trinary-trytes.md
samuel-rufi: Agree @thibault-martinez and @mikrohash, this would contribute to correctness and safety.

text/0000-trinary-trytes/0000-trinary-trytes.md
```diff
@@ -0,0 +1,117 @@
...
+Type `Trytes` is a type `BytesMut` and moreover a smart pointer. There are also
+a series of fixed sized type to help checking at compiled time to increase
+correctness and performance.
+The structure simply looks like this:
+
+```rust
+pub type Tryte = u8;
+
+pub type Tag = [Tryte; 9];
+pub type Adress = [Tryte; 81];
```
thibault-martinez, line 48: Should we define transaction fields in terms of trytes or in terms of trits ? And anyway they shouldn't be defined here but in the model crate. What we can do here is just define the usual sizes: `Tryte9`, `Tryte27`...

SuperFluffy:
At the moment it is not clear at all what the use case of the `trytes` type is supposed to be. The only motivating argument brought forward for its existence is the demarcation from the `trits`, quote:

> The reason to implement trytes only at the moment is not only to be loosly coupled from **trits**, but also layout what a ternary type could looks like.

I summarize the discussion that we had between @Alex6323, @mikrohash, @rufsam, @wusyong, and @SuperFluffy:

> The primary use for an encoding that is a 1-to-1 from trytes to bytes is for the purposes of debugging and eventual display to stdout. For the purpose of data intake and manipulation (so, data coming in from the tangle, verification, hashing, signing, etc), 2-to-9 or 2-to-10 (tryte-to-byte) encodings are the norm.

This implies that the 1-to-1 encoding does not appear to be elemental for core functionality at this point. Furthermore, if the purpose of the `1-to-1` encoding is to only display it to the end user, we might just implement the `Debug` and/or `Display` traits by decoding the 2-to-9/10 encodings on the fly, i.e. without allocating a `Trytes` object.

Based on this, I would thus vote to postpone the implementation of this feature until such a time whence we consider user facing APIs. Instead, we should focus on the more fundamental 2-to-9/10 encodings and then reconsider how this 1-to-1 can interact with those.

thibault-martinez:
@SuperFluffy 

> At the moment it is not clear at all what the use case of the trytes type is supposed to be. 

That is not entirely true. We already know *for sure* that we are going to deal with raw trytes in the  following  situations.
- API
- Configuration
- Spent addresses files
- Snapshot files
- Tests

The question now is do we want to consider them as strings or do we want to have a proper type ? I'll always vouch for a proper type, even typed on the size if possible.

I agree this RFC is less pressing than the trits one, but at the end of the day, it's only a typedef.

text/0000-trinary-trytes/0000-trinary-trytes.md
wusyong: Sure, the definition can change. I just give some examples. Will update to this later.

SuperFluffy:
@thibault-martinez 

Thank you for the clarification. That indeed sounds like a proper type will be useful.

The scenarios you mentioned should definitely be mentioned in the motivation for this this RFC. More importantly than that, since this is (end) user facing API which you cannot change as easily as internal implementation details, much more care has to be taken when designing this, including the appropriate bike shedding.

SuperFluffy:
The team decided to postpone this feature for now and revisit it until the core functionality is implemented and until we start focusing on user-facing APIs.

Issue #5, opened 2019-09-11
pre-RFC: suggestion of Trits implementation


wusyong:
@iotaledger/bee-devs With some conversation on #4, we should better start figuring out the basic layout of type `Trits` and open its RFC at the meantime. Here's proposal of @zesterer in trit's manner:
```rust
pub trait RawTrits {
    fn with_capacity(n: usize) -> Self;
    fn double_in_place(&mut self) -> bool;
    fn reserve(&mut self, used: usize, needed: usize);
    ...
}

pub struct RawTritPerByte {
    trits: ...,
} 

impl RawTrits for RawTritPerByte {

}

pub struct Trits<T: RawTrits>{
    raw: T,
}

impl<T: RawTrits> Trits<T> {
    pub fn new() -> Self {
        ...
    }
    
    pub fn with_capacity(n: usize) -> Self {
        ...
    }
    
    pub fn push(&mut self, x: Trits) {
        ...
    }
    
    pub fn pop(&mut self) -> Option<Trits> {
        ...
    }
}
```

So it can have default type something like `struct Trits<T: RawTrits = RawTritPerByte>` in this way. We still have a couple of questions remain and I would like to see if some one wish to take it and start a RFC with this.

- What trits encoding are we going to start first?
- Or should we implement both at once?
- How to do hashing with different encoding
- What method should we provide at least to defin in trait `RawTrits`?

zesterer:
> How to do hashing with different encoding

Hashing can be implemented on top of the high-level API (i.e: on the `Trits` struct). Alternatively, it could be implemented as a method on the `RawTrits` trait, with a default (but potentially slower) implementation provided. This way, encodings can override the default implementation if they have a more performant algorithm.

thibault-martinez:
Some thoughts about this.

It seems like we are aiming at supporting the following trit encodings.
- `1t/1b` - Raw trits
- `5t/1b` - Mainnet wire
- `9t/2b` - Potential Bee internal encoding
- `3t/1b` - Trytes (the trytes type could actually just be a typedef on this)
- `27t/8b` - Troika optimization
- `ptrit` - A little bit more challenging but makes sense

They should all share a common interface and should provide a way to convert from one to another.
This can easily be achieved by providing a `toRaw` and `fromRaw` methods.

thibault-martinez:
> 3t/1b - Trytes (the trytes type could actually just be a typedef on this)

[...]

PR #10, opened 2019-09-20
RFC: `trits` module

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/rfc-bee-trit/text/0010-trits-module/0010-trits-module.md)

Based on https://github.com/iotaledger/bee-rfcs/issues/5


text/0000-trits/0000-trits.md
```diff
@@ -0,0 +1,95 @@
...
+| `ptrit`   | Hardware dependant    | PCurl batch hashing   |   |
+
+Because there are many different trits encodings and switching from one to another is often needed, having a trits interface to have a uniform way to interact with trits, whatever the underlying encoding, could be useful.
+
+This interface should offer methods like allocations, conversions and manipulations.
+
+# Detailed design
+
+```rust
+pub trait Trinary {
```
wusyong: I renamed this to `Trinary` hope this can be more clear as a trait aspect.

text/0000-trits/0000-trits.md
```diff
@@ -0,0 +1,95 @@
...
+    fn with_capacity(n: usize) -> Self;
+    fn double_in_place(&mut self) -> bool;
+    fn reserve(&mut self, used: usize, needed: usize);
+    fn to_trits(ref: &[u8]) -> Self;
+    ...
+}
+
+pub struct 1T1B {
+    raw: ...,
```
wusyong: I name each encoded type's field called `raw` and `inner` for `Trits`. Hope this can be more specific.

text/0000-trits/0000-trits.md
```diff
@@ -0,0 +1,95 @@
...
+    pub fn push(&mut self, x: Trits) {
+        ...
+    }
+    
+    pub fn pop(&mut self) -> Option<Trits> {
+        ...
+    }
+}
+
+impl<T: Trinary> From<[u8]> for Trits<T> {
```
wusyong: Ideally, each Trits type can implement any `From<T>` it wants, and should have one for `1T1B` at least.

text/0000-trits/0000-trits.md
```diff
@@ -0,0 +1,95 @@
...
+
+# Summary
+
+WIP
+
+# Motivation
+
+A `trit` is a digit in base `3` and is usually represented by either `-1`, `0` or `1`.
+
+Even though the most convenient way to encode trits is to map one `trit` to one `byte`, this is obviously not the most memory efficient way nor the most visual way to do it.
```
tsvisabo: visual => readable

text/0000-trits/0000-trits.md
samuel-rufi: In the case I want to get the bytes from the trits (e.g. to send them through the network), we also need a function for that; or what is the idea?

thibault-martinez: @rufsam Have a look at https://github.com/iotaledger/bee-rfcs/pull/10/commits/97805c0e580a6e87cc55b2ec23abaa5ace6458d9

thibault-martinez: I think it makes more sense to have a `to` than a `from` 🤔

text/0000-trits/0000-trits.md
```diff
@@ -0,0 +1,104 @@
...
+
+This interface should offer methods like allocations, conversions and manipulations.
+
+# Detailed design
+
+Any encoding should be convertible to any encoding. If `N` is the number of encodings available, then the number of possible conversions is `N^2` which could grow quickly. While some of these conversions are heavily used (e.g. `5t/1b` to/from `9t/2b`) and need to be as efficient as possible, some other are rarely or never expected to be used (e.g. `5t/1b` to/from `3t/1b`).
+
+To allow implementing all conversions without having to focus too much on unnecessary work, we propose the following requirements:
+- a conversion from any encoding to the same encoding is done by copy;
+- a conversion from `1t/1b` to any other encoding is efficiently implemented;
```
samuel-rufi: Why use 1t/1b as conversion between two different encodings, and not any other one?

text/0000-trits/0000-trits.md
```diff
@@ -36,6 +36,15 @@ This interface should offer methods like allocations, conversions and manipulati
 
 # Detailed design
 
+Any encoding should be convertible to any encoding. If `N` is the number of encodings available, then the number of possible conversions is `N^2` which could grow quickly. While some of these conversions are heavily used (e.g. `5t/1b` to/from `9t/2b`) and need to be as efficient as possible, some other are rarely or never expected to be used (e.g. `5t/1b` to/from `3t/1b`).
+
+To allow implementing all conversions without having to focus too much on unnecessary work, we propose the following requirements:
+- a conversion from any encoding to the same encoding is done by copy;
+- a conversion from `1t/1b` to any other encoding is efficiently implemented;
+- a conversion from any encoding to `1t/1b` is efficiently implemented;
+- a conversion, proven to be heavily used, from any encoding to any other encoding is efficiently implemented;
```
zesterer: I wonder whether it's worth looking into whether Rust's specialisation support would make this even nicer to implement?

text/0000-trits/0000-trits.md
thibault-martinez: Since hashing happens on `1t/1b`, efficient conversion from/to it should already be efficiently implemented so it makes it a good intermediate candidate.

text/0000-trits/0000-trits.md
wusyong: This is just for the usage of `from`. There's also a method `to_trits` in trait `Trinary` which can also be provided method in the `Trits<T>` interface.

wusyong: I would like to also mention that `into` is also defined once `From` trait is implemented. Same for `TryFrom`.

bingyanglin: This is a good ref: https://doc.rust-lang.org/rust-by-example/conversion/from_into.html

wusyong (approved):

text/0000-trits/0000-trits.md
```diff
@@ -0,0 +1,106 @@
...
+- any other conversion falls back to a trivial conversion with an intermediary `1t/1b` step;
+
+```rust
+pub trait Trinary {
+    fn with_capacity(n: usize) -> Self;
+    fn to_trits(ref: &[u8]) -> Self;
```
bingyanglin: It is good to follow the Rust type casting convention, so we have 
`impl<T: Trinary> From<[u8]> for Trits<T>` only, and the `Into` is automatically yielded.
We don't need `fn to_trits(ref: &[u8]) -> Self;`

[This discussion](https://stackoverflow.com/questions/29812530/when-should-i-implement-stdconvertfrom-vs-stdconvertinto) provides a good thought.

wusyong: This is the inner structure for `Trits` to use. We don't want this type to expose to public, so there's only selective methods to use. Traits like `From` should also call this to do the conversion.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,106 @@
...
+
+At the time this RFC is written, the following encodings are considered. This list is not exhaustive, some may be dropped, some may be added.
+
+| Name      | Encoding              | Usage                 | RFC |
+| --------- | --------------------- | --------------------- | - |
+| `t1/b1`   | 1 trit per byte       | Hashing, tests        |   |
+| `t3/b1`   | 3 trits per byte      | API, UX, tests        |   |
+| `t27/b8`  | 27 trits per 8 bytes  | Troika optimization   |   |
+| `t9/b2`   | 9 trits per 2 bytes   | Bee internal encoding |   |
+| `t5/b1`   | 5 trits per byte      | Network communication |   |
```
thibault-martinez: + storage

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,121 @@
...
++ RFC PR: [iotaledger/bee-rfcs#10](https://github.com/iotaledger/bee-rfcs/pull/10)
++ Bee issue: [iotaledger/bee#45](https://github.com/iotaledger/bee/issues/45)
+
+# Summary
+
+This RFC proposes a `trits` interface to standardise the usage of different underlying encodings and their conversions from one to another.
+
+# Motivation
+
+A `trit` is a digit in base `3` and is usually represented by either `-1`, `0` or `1`.
```
Alex6323: Should we be more precise here and write? "A balanced ternary `trit` is a digit ...."

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,121 @@
...
+
+Even though the most convenient way to encode trits is to map one `trit` to one `byte`, this is obviously not the most memory efficient way nor the most readable way to do it.
```
Alex6323: more precisely: one balanced `trit` to one `signed byte`

text/0010-trits/0010-trits.md
thibault-martinez: Can't hurt to be more precise :)

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
++ RFC PR: [iotaledger/bee-rfcs#10](https://github.com/iotaledger/bee-rfcs/pull/10)
++ Bee issue: [iotaledger/bee#45](https://github.com/iotaledger/bee/issues/45)
+
+# Summary
+
+This RFC proposes a `trits` interface to standardise the usage of different underlying encodings and their conversions from one to another.
+
+# Motivation
+
+A `trit` is a digit in base `3` and is usually represented, in balanced ternary, by either `-1`, `0` or `1`.
```
SuperFluffy: I think the following wording is more precise:

> A `trit` is a sigit in base 3 and is often represented by the set `{-1, 0, +1}`. This representation is called *balanced ternary* and the object represented in such a way a `balanced trit`.

We might also put a reference here for further reading: http://homepage.divms.uiowa.edu/~jones/ternary/bct.shtml

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+
+Even though the most convenient way to encode trits is to map one balanced `trit` to one `signed byte`, this is obviously not the most memory efficient way nor the most readable way to do it.
```
SuperFluffy: To work with `trits` on a binary machine, the most apparent solution is to encode them with the mapping `1 balanced trit -> one signed byte`. This, however, is not very memory efficient, because a byte can represent 256 different values, while a trit only represents 3.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+
+In the IOTA ecosystem, different trits encodings are used in different situations like:
```
SuperFluffy: (The struckthrough is hard to read, I changed `trits -> trit`)

> In the IOTA ecosystem, different trit~s~ encodings are used in different situations like:

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+At the time this RFC is written, the following encodings are considered. This list is not exhaustive, some may be dropped, some may be added.
+
+| Name    | Encoding              | Usage                           | RFC |
+| ------- | --------------------- | ------------------------------- | --- |
+| `t1b1`  | 1 trit per byte       | Hashing, tests                  |     |
+| `t3b1`  | 3 trits per byte      | API, UX, tests                  |     |
+| `t27b8` | 27 trits per 8 bytes  | Troika optimization             |     |
+| `t9b2`  | 9 trits per 2 bytes   | Bee internal encoding           |     |
+| `t5b1`  | 5 trits per byte      | Network communication, storage  |     |
+| `ptrit` | Hardware dependant    | PCurl batch hashing             |     |
```
SuperFluffy: I would like to mention what the `p` in `ptrit` stands for.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+
+Because there are many different trits encodings and switching from one to another is often needed, having a trits interface to have a uniform way to interact with trits, whatever the underlying encoding, could be useful.
```
SuperFluffy: Better:

> Because there are many different trits encodings and switching from one to another is needed often, defining a trits interface to have a uniform way to interact with trits, whatever the underlying encoding, could be useful.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+
+This interface should offer methods like allocations, conversions and manipulations.
+
+# Detailed design
+
+Any encoding should be convertible to any encoding. If `N` is the number of encodings available, then the number of possible conversions is `N^2` which could grow quickly. While some of these conversions are heavily used (e.g. `t5b1` to/from `t9b2`) and need to be as efficient as possible, some other are rarely or never expected to be used (e.g. `t5b1` to/from `t3b1`).
```
SuperFluffy: It's `N * (N-1)`, isn't it? Saying “it grows like `N^2`” would probably be more correct.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+
+To allow implementing all conversions without having to focus too much on unnecessary work, we propose the following requirements:
```
SuperFluffy: It's indeed important to provide conversion between the various encodings. However, do we need to specify in so much detail what these are supposed to be? Also, the importance might change in the future, and it might also be different for different users.

Let's leave it as an implementation detail how exactly the conversion should take place, and only note the conversion between encodings is required.

Less important: linebreak before starting the list please.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+- a conversion from any encoding to the same encoding is done by copy;
```
SuperFluffy: The right term is `clone` in Rust terminology. You mean a deepcopy, right?

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+
+- hashing is done on raw trits because IOTA uses ternary hash functions like `Curl` or `Troika`;
+- sending trits on the network is done by encoding `5` trits to a single byte for throughput efficiency;
+- for UX purposes, we usually encode `3` trits to a single byte because of the nice readable mapping it offers (`9ABCDEFGHIJKLMNOPQRSTUVWXYZ`);
+
+At the time this RFC is written, the following encodings are considered. This list is not exhaustive, some may be dropped, some may be added.
+
+| Name    | Encoding              | Usage                           | RFC |
+| ------- | --------------------- | ------------------------------- | --- |
+| `t1b1`  | 1 trit per byte       | Hashing, tests                  |     |
```
SuperFluffy: Let's mention explicitly that by the naming convention `tmbn` we mean `m trits represented by n bytes`.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
++ Feature name: trits
```
Alex6323: `trits` seems a bit too generic. It's hard to guess from it what this RFC is about.
Suggestions: `trits-api`, or `trits-interface`, or `trits-module`

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    pub fn len(&self) -> usize {
+        self.inner.len()
+    }
+
+    ...
+}
+
+impl<T: Encoding> Deref for Trits<T> {
```
SuperFluffy: This exposes the implementation detail to the outside. I would not do this.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+
+Note that inner structure should not provide any method to public. It should be
+implemented by methods of `Trits<T>` to that specific type. This interface can
+dereference to inner structure and provide more functionality other than general
+methods.
+
+```rust
+pub trait Encoding {
+    fn with_capacity(n: usize) -> Self;
+    fn len(&self) -> usize;
```
SuperFluffy: `len` and `with_capacity` above suggest that we will be using variable sized datastructures underlying the various representations. Is that intended? Isn't this a core design decision that should be made explicit?

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+- a conversion, proven to be heavily used, from any encoding to any other encoding is efficiently implemented by `From` trait with specialization support;
+- any other conversion falls back to a trivial conversion with an intermediary `t1b1` step;
+
+Note that inner structure should not provide any method to public. It should be
+implemented by methods of `Trits<T>` to that specific type. This interface can
+dereference to inner structure and provide more functionality other than general
+methods.
+
+```rust
+pub trait Encoding {
```
SuperFluffy: I would probably give this a more explicit name, like `TritsAsBytes` or `BinaryCodedTernary`.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+
+pub struct T1B1 {
+    raw: ...,
+}
+
+impl Encoding for T1B1 {
+    ...
+}
+
+pub struct Trits<T: Encoding>{
```
SuperFluffy: No need to have the trait bound at this point.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
++ Feature name: trits
++ Start date: 2019-09-20
++ RFC PR: [iotaledger/bee-rfcs#10](https://github.com/iotaledger/bee-rfcs/pull/10)
++ Bee issue: [iotaledger/bee#45](https://github.com/iotaledger/bee/issues/45)
+
+# Summary
+
+This RFC proposes a `trits` interface to standardise the usage of different underlying encodings and their conversions from one to another.
```
Alex6323: Suggestion: Maybe it would make sense to add some additional context here or in the motivation section why we pursue a standardized interface here.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+9T2B are not required because they are mainly used for transfer.
+
+We can implement more methods but it still depends on what trait `Encoding` can
+provide. Or we give more freedom to have method like `into_inner` to manipulate
+the inner structure directly, but this might contradict to the purpose of this
+interface.
+
+# Rationale and alternatives
+
+- Just define type alias and some convert functions. This give way more freedom to the users.
```
SuperFluffy: A `type alias` is probably not what is meant here, but `newtype` instead. A type alias would still expose implementation details because it's just a different name for another type.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+- for UX purposes, we usually encode `3` trits to a single byte because of the nice readable mapping it offers (`9ABCDEFGHIJKLMNOPQRSTUVWXYZ`);
+
+At the time this RFC is written, the following encodings are considered. This list is not exhaustive, some may be dropped, some may be added.
+
+| Name    | Encoding              | Usage                           | RFC |
+| ------- | --------------------- | ------------------------------- | --- |
+| `t1b1`  | 1 trit per byte       | Hashing, tests                  |     |
+| `t3b1`  | 3 trits per byte      | API, UX, tests                  |     |
+| `t27b8` | 27 trits per 8 bytes  | Troika optimization             |     |
+| `t9b2`  | 9 trits per 2 bytes   | Bee internal encoding           |     |
```
Alex6323: It is debatable, but I think we should consider `t9b2` for network communication among Bee nodes as well, because that would release the nodes from having to convert between `t5b1` and `t9b2` in those cases. For (maybe battery operated) smaller devices it is probably more beneficial to favor CPU load reduction over bandwidth/memory consumption.

text/0010-trits/0010-trits.md
Alex6323: I think this is the sentence I missed to read further above. I don't see any particular reason, why it couldn't be moved after line 8 as it supports the argument there.

text/0010-trits/0010-trits.md
Alex6323: Do we really expect many more encodings to come?

text/0010-trits/0010-trits.md
Alex6323: Is there a usecase for this?

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+- any other conversion falls back to a trivial conversion with an intermediary `t1b1` step;
+
+Note that inner structure should not provide any method to public. It should be
+implemented by methods of `Trits<T>` to that specific type. This interface can
+dereference to inner structure and provide more functionality other than general
+methods.
+
+```rust
+pub trait Encoding {
+    fn with_capacity(n: usize) -> Self;
```
Alex6323: the argument should be named idiomatically 'capacity', not just 'n'.
if you add `with_capacity`, wouldn't it make sense to also add the function `fn new() -> Self` to the trait in case the capacity is unknown?

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+    fn len(&self) -> usize;
+    ...
```
Alex6323: How can we add data to the internal datastructure? What is supposed to replace the blank `...`?

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+}
+
+pub struct Trits<T: Encoding>{
+    inner: T,
+}
+
+impl<T: Encoding> Trits<T> {
+    pub fn new() -> Self {
+        Self { inner: T::with_capacity(0) }
```
Alex6323: why not just let the trait implementor have a constructor 'new'?

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+    ...
+}
+
+impl<T: Encoding> Deref for Trits<T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
```
Alex6323: So we want to hide T behind Trits, but here we admit that we actually can't? I'm not sure about whether this is the right approach then. At the surface it seems we could basically swap each encoding for another, and operate the same way with them, but that essentially means an operation like adding a single trit to each encoding would be a semantically valid operation. Is it? It's certainly fine for T1B1, but is it still for T3B1 (easy fix: just magically assume the API user wants the remaining two trits to be set to 0) and all the other encodings? Not saying we shouldn't do this. Just that the proposal should mention those cases and how they will be handled.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+interface.
+
+# Rationale and alternatives
+
+- Just define type alias and some convert functions. This give way more freedom to the users.
+- Each encoded trit is a type of its own instead of interface like above.
+
+# Unresolved questions
+
+- Do we want to type usual sizes for more safety ?
```
Alex6323: definitely makes sense for fully decoding a transaction in a client lib where memory savings are not a major concern. The Bee node however shouldn't do that (full tx decoding, i.e. allocating memory for each transaction field) for efficiency reasons.

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+- In case of missing conversion, how are fallbacks to slower implementation handled ?
```
Alex6323: IMHO this can be handled by the API user. He realizes a direct conversion is not supported, so he simply has to do 2 conversion steps (source_enc -> bridge_enc -> target_enc)

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+- Which type we should to do first on the initial implementation ?
```
Alex6323: the order doesn't really matter because we need all those basic encodings just for development/debugging/testing purposes alone (not talking about `ptrit` encoding which can be done later).

text/0010-trits/0010-trits.md
```diff
@@ -0,0 +1,124 @@
...
+- Should we do all conversion once and for all, or just provide minimum ways to do it first ?
```
Alex6323: I don't understand what the question is here tbh. Can you elaborate? Do you mean all conversion between all encodings?

text/0010-trits/0010-trits.md
thibault-martinez: I wanted to rename it to `trits-module` as well.

text/0010-trits/0010-trits.md
thibault-martinez: I believe this is actually explained in Motivation ?

text/0010-trits/0010-trits.md
thibault-martinez: I definitely agree, but this table is an overview of the existing state of encodings. These discussions will happen in later RFCs.

text/0010-trits/0010-trits.md
thibault-martinez: I feel like summary should give a short insight of what's to come. This sentence is definitely motivation-related IMHO.

text/0010-trits/0010-trits.md
thibault-martinez: Who knows :) ?
But to be serious, other  than cryptic hash optimizations I don't think so.

text/0010-trits/0010-trits.md
thibault-martinez: I'm fine with that for lack of better alternative.

text/0010-trits/0010-trits.md
thibault-martinez: Yeah not sure about this question, I'd remove it.

text/0010-trits/0010-trits.md
thibault-martinez: I don't understand it either. @wusyong can you elaborate ?

thibault-martinez: I guess he was referring to us providing the most common conversions from one encoding to another, and letting the rest fall back to 2 step with T1B1

text/0010-trits/0010-trits.md
thibault-martinez: Other than just allowing it to compile, I don't think so.

text/0010-trits/0010-trits.md
thibault-martinez: For the record, we agreed to have that later in another RFC.

text/0010-trits/0010-trits.md
Alex6323: aha, okay

SuperFluffy (changes requested):
The motivation for a representing ternary numbers is not completely worked out yet, especially not the aspect why conversion between the various encodings is necessary. Reading the RFC, I was asking myself:

a) How do these encodings enter a running program?
b) Under which circumstances is conversion between the encodings required?
c) What does `bee internal encoding` mean? When a ternary coded message enters a program, shouldn't it be decoded into some internal memory representation (for example, a `struct` with fields matching the parts of the IOTA wire protocol)?
d) Will `bee` ever deal with data that is kept in `ternary format` for the entire lifetime of that data?

Especially with respect to point c) I am thinking that a natural data flow would be `messages comes in -> message is decoded into struct -> data is manipulated -> data is encoded into ternary message`.

Furthermore, while I understand the need for a `trits` module (note: I am in favour of the name `bee-ternary`) given the protocols on which IOTA run, I do not understand the justification for the `Encoded` trait in its current form.

You can get a comparable behavior of all the different encodings just by convention, i.e. by defining `with_capacity`, `len`, and other functions as methods on the different types. It does not become clear from the proposal how the common interface as defined by the `Encoded` trait will help with defining conversions between the types.

Also, I would always opt for hiding the details of how we choose to implement the various encodings to be flexible in the future. At the moment, the suggested `Deref` trait exposes this implementation detail (the underlying data structure to store trits).

text/0010-trits/0010-trits.md
wusyong: This is why we going to have other encoding like t9b2 or t5b1

thibault-martinez: I think @SuperFluffy was just rephrasing the sentence :)

text/0010-trits/0010-trits.md
wusyong: It was EncodedTrits previously. Do you wish to roll back?

SuperFluffy: I am not a fan of `EncodedTrits` either, because it doesn't convey what we are actually doing. We can also leave naming/bikeshedding to the implementation phase.

text/0010-trits/0010-trits.md
SuperFluffy: Yeah, I forgot to add that I was rephrasing. :)

text/0010-trits/0010-trits.md
wusyong: I'm not sure yet tbh. This is why I `impl Deref` in case of trait methods are not enough.

text/0010-trits/0010-trits.md
wusyong: I'm going to also answer question above. Trait should have methods that really necessary. If this can complete `new` and `with_capacity` at the same time. Why should we have redundant method in trait?

text/0010-trits/0010-trits.md
wusyong: If we don't implement Deref, we should have well provided methods in trait especially for conversion.

wusyong: Also note that if you want to add trits, it should implement `DerefMut` instead.

text/0010-trits/0010-trits.md
wusyong: Then we should propose well-provided methods in `Encoding` traits.

text/0010-trits/0010-trits.md
wusyong: Yeah, we have a discussion yesterday. We might change this to fit for fixed-size structure.

text/0010-trits-module/0010-trits-module.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Motivation
+
+A `trit` is a digit in base `3` and is often represented by the set `{-1, 0, +1}`. This representation is called balanced ternary and the object represented in such a way a `balanced trit`. Further reading: http://homepage.divms.uiowa.edu/~jones/ternary/bct.shtml.
+
+To work with trits on a binary machine, the most apparent solution is to encode them with the mapping `one balanced trit -> one signed byte`. This, however, is not very memory efficient, because a `byte` can represent `256` different values, while a `trit` only represents `3`.
+
+In the IOTA ecosystem, different trit encodings are used in different situations like:
+
+- hashing is done on raw trits because IOTA uses ternary hash functions like `Curl` or `Troika`;
```
rufsam: Maybe explain the difference between raw trits and non raw trits.

samuel-rufi, line 18: Maybe explain what is meant with `raw` trits. A reader might ask if there exists something like non-raw trits and if yes how it differs.

text/0010-trits-module/0010-trits-module.md
```diff
@@ -0,0 +1,125 @@
...
+
+# Summary
+
+This RFC proposes a `trits` interface to standardise the usage of different underlying encodings and their conversions from one to another.
+
+# Motivation
+
+A `trit` is a digit in base `3` and is often represented by the set `{-1, 0, +1}`. This representation is called balanced ternary and the object represented in such a way a `balanced trit`. Further reading: http://homepage.divms.uiowa.edu/~jones/ternary/bct.shtml.
+
+To work with trits on a binary machine, the most apparent solution is to encode them with the mapping `one balanced trit -> one signed byte`. This, however, is not very memory efficient, because a `byte` can represent `256` different values, while a `trit` only represents `3`.
```
oopsmonk, line 14:
```suggestion
To work with trits on a binary machine, the most apparent solution is to encode them with the mapping `one balanced trit -> one signed byte`. This, however, is not very memory efficient, because a `byte` can represent `256` different values and is able to store multiple trits through encoding methods.
```

tsvisabo (approved):

PR #13, opened 2019-09-25
RFC: `trits-t5b1` encoding

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/rfc-t5b1/text/0013-trits-t5b1/0013-trits-t5b1.md)


text/0000-trits-t5b1/0000-trits-t5b1.md
```diff
@@ -0,0 +1,119 @@
...
+
+A `byte` is composed of `8` `bits` that can hold `2^8 = 256` different values. On the other hand, `6` `trits` can hold `3^6 = 729` values while `5` `trits` can hold `3^5 = 243` values. The maximum number of `trits` a `byte` can hold is then `5`.
+
+As this is the most memory-efficient way to encode `trits` in `bytes`, `t5b1` is often used for network communications to increase throughput and for storage layers to reduce memory consumption.
+
+# Detailed design
+
+For efficient conversions, we can make use of a LookUp Table (`LUT`).
+
+We give the `t5b1` `LUT` as a reference for implementation, it contains the `243` possible values ordered like this: `0`, `1`, ... `120`, `121`, `-121`, `-120`, ... `-2`, `-1`.
```
semenov-vladyslav: The `LUT` should cover all `256` possible cases for values of a byte. To the byte values in range `243..255` we can assign some default mapping to 5 trits, eg. 5 zero trits (ie. 5 zero trits can be encoded in several ways). This way we can avoid runtime checks (branching on each iteration of a loop can degrade performance in case of false branch prediction). Anyway, all data checks must be cryptographic, ie. using hash, MAC or sig. 

Additionally, we can have two variants of "decode from bytes" function: 1. checked & slow (checks for correct values for each byte and returns false in case of incorrect value) and, 2. fast & unchecked.

The overhead of 13 additional values is not very big: +~5% to the size of `LUT`.

text/0000-trits-t5b1/0000-trits-t5b1.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Encoding
+
+<!-- TODO -->
+
+## Decoding
+
+### With LUT
+
+Decoding a `byte` into `5` `trits` with the help of the `LUT` is fairly simple:
```
semenov-vladyslav: How should decoding behave when we need to decode a `byte` into less than `5` trits (for example, for 27 encoded trits there are only 2 trits encoded in the tail byte)? Just decode `byte` into `5` trits and drop unneeded trits?

thibault-martinez: You can encode just 2 trits to a byte,  just consider the 3 other trits as 0

text/0000-trits-t5b1/0000-trits-t5b1.md
thibault-martinez: Good point!

text/0000-trits-t5b1/0000-trits-t5b1.md
```diff
@@ -0,0 +1,119 @@
...
+    [ 1, 0, -1, 0, 0 ],    [ -1, 1, -1, 0, 0 ],   [ 0, 1, -1, 0, 0 ],     [ 1, 1, -1, 0, 0 ],    [ -1, -1, 0, 0, 0 ],
+    [ 0, -1, 0, 0, 0 ],    [ 1, -1, 0, 0, 0 ],    [ -1, 0, 0, 0, 0 ]
+]
+```
+
+Because the memory footprint (`1.2kb`) of this `LUT` can be too high for some very limited systems, we also present algorithms that do not use it.
+
+## Encoding
+
+<!-- TODO -->
```
semenov-vladyslav: Although trit variables should contain only `-1`, `0` or `1` they are however of type `i8` and these restrictions are not forced by compiler. Invalid out of range trit values can also come from unchecked javascript code. So there are two solutions to the problem: 1. make encoding checked and fail if some trit is out of range, but it's rather slow, and 2. make encoding fail-safe by assigning some default value to invalid values of a trit (it'll be safe and reasonably fast).

A fail-safe and fast encoding can look like this: 
```rust
LUT: [[[[[u8; 4]; 4]; 4]; 4]; 4];
fn encode_unchecked(t: &[i8; 5]) -> u8 {
  let i0 = t[0] & 3; // take lower two bits
  let i1 = t[1] & 3;
  let i2 = t[2] & 3;
  let i3 = t[3] & 3;
  let i4 = t[4] & 3;
  LUT[i0][i1][i2][i3][i4]
}
```

In this case we have the following correspondence between values of type `i8` and trits:
`0->0`, `1->+1`, `2->0`, `3->-1`, `4->0`, `5->+1`, `6->0`, `7->-1`, and so on.

In addition to a fast unchecked (but safe) function we can have a checked (but slower) function.

thibault-martinez: Since this is implementation detail I think I'll post the raw LUTs with no modification to them but insist on the safety concerns by including what you just wrote.

text/0000-trits-t5b1/0000-trits-t5b1.md
```diff
@@ -0,0 +1,119 @@
...
+- `6561` `trits` fit in `1313` `bytes`
+- `8019` `trits` fit in `1604` `bytes`
+
+# Drawbacks
+
+<!-- TODO -->
+
+# Rationale and alternatives
+
+<!-- TODO -->
```
semenov-vladyslav: What other operations should be supported for an encoding? If we consider `[u8]` as array of trit quintuples we might want to copy some subrange of trits that is not aligned to a bound of 5 trits (for example, copy trits from 2 to 13 encoded as bytes): 
```
fn copy(from: &[u8], from_offset: usize, to: &mut [u8], to_offset: usize, count: usize)
```
A trivial implementation would first decode `from` into trits and then encode subrange of the trits into `to` bytes. But it uses additional memory and two conversions (from and to bytes). A more efficient implementation would use no additional memory and one conversion via a more intricate LUT.

text/0013-trits-t5b1/0013-trits-t5b1.md
```diff
@@ -0,0 +1,322 @@
...
+    [ 1, 0, -1, 0, 0 ],    [ -1, 1, -1, 0, 0 ],   [ 0, 1, -1, 0, 0 ],     [ 1, 1, -1, 0, 0 ],    [ -1, -1, 0, 0, 0 ],
+    [ 0, -1, 0, 0, 0 ],    [ 1, -1, 0, 0, 0 ],    [ -1, 0, 0, 0, 0 ]
+]
+```
+
+Decoding a `signed byte` into `5` `balanced` `trits` with the help of the `LUT` is fairly simple:
+- `byte >= 0`: use `byte` as an index for the `LUT` e.g. if the `byte = 42` then `LUT[42] = [ 0, -1, -1, -1, 1 ]` and `0 * 3^0 + (-1) * 3^1 + (-1) * 3^2 + (-1) * 3^3 + 1 * 3^4 = 42`;
+- `byte < 0`: use `byte + 243` as an index for the `LUT` e.g. if the `byte = -42` then `LUT[201] = [ 0, 1, 1, 1, -1 ]` and `0 * 3^0 + 1 * 3^1 + 1 * 3^2 + 1 * 3^3 + (-1) * 3^4 = -42`;
+
+Even though this RFC won't cover every little implementation details, it can be noted that the `LUT` could cover all `256` possible values of a `byte` to avoid invalid accesses. To the missing `byte` values we can simply assign some default mapping to `5` `trits`, eg. `[ 0, 0, 0, 0, 0 ]`. This way runtime checks can be avoided as branching on each iteration of a loop can degrade performance in case of false branch prediction.
```
tsvisabo: details => detail

text/0013-trits-t5b1/0013-trits-t5b1.md
```diff
@@ -0,0 +1,322 @@
...
+
+Additionally, there could be two variants of the decoding function:
+- checked & slow;
+- unchecked & fast;
+
+The overhead of the `13` additional values is only +~5% to the size of `LUT`.
+
+### Without LUT
```
tsvisabo, line 96: taking into account functions that get/set one trit in this encoding (or any other) without LUT 
it would take nasty bitwise operations, we should think about it if we even want to allow anything without LUT,
IMO, this is a really tiny overhead  considering the Ledger itself, so, i struggle to find why it should be only optional

tsvisabo (changes requested):

text/0013-trits-t5b1/0013-trits-t5b1.md
thibault-martinez: The size of the LUTs is non negligible for some very limited devices. Also, having two different implementations for testing purposes is great.

tsvisabo (approved):

text/0013-trits-t5b1/0013-trits-t5b1.md
```diff
@@ -0,0 +1,324 @@
...
+    [ -1, 0, 0, -1, 0 ],   [ 0, 0, 0, -1, 0 ],    [ 1, 0, 0, -1, 0 ],     [ -1, 1, 0, -1, 0 ],   [ 0, 1, 0, -1, 0 ],
+    [ 1, 1, 0, -1, 0 ],    [ -1, -1, 1, -1, 0 ],  [ 0, -1, 1, -1, 0 ],    [ 1, -1, 1, -1, 0 ],   [ -1, 0, 1, -1, 0 ],
+    [ 0, 0, 1, -1, 0 ],    [ 1, 0, 1, -1, 0 ],    [ -1, 1, 1, -1, 0 ],    [ 0, 1, 1, -1, 0 ],    [ 1, 1, 1, -1, 0 ],
+    [ -1, -1, -1, 0, 0 ],  [ 0, -1, -1, 0, 0 ],   [ 1, -1, -1, 0, 0 ],    [ -1, 0, -1, 0, 0 ],   [ 0, 0, -1, 0, 0 ],
+    [ 1, 0, -1, 0, 0 ],    [ -1, 1, -1, 0, 0 ],   [ 0, 1, -1, 0, 0 ],     [ 1, 1, -1, 0, 0 ],    [ -1, -1, 0, 0, 0 ],
+    [ 0, -1, 0, 0, 0 ],    [ 1, -1, 0, 0, 0 ],    [ -1, 0, 0, 0, 0 ]
+]
+```
+
+Decoding a `signed byte` into `5` `balanced` `trits` with the help of the `LUT` is fairly simple:
```
wusyong, line 84: Perhaps also mention that a `signed byte` is `i8` in rust.

text/0013-trits-t5b1/0013-trits-t5b1.md
```diff
@@ -0,0 +1,324 @@
...
+- `byte >= 0`: use `byte` as an index for the `LUT` e.g. if the `byte = 42` then `LUT[42] = [ 0, -1, -1, -1, 1 ]` and `0 * 3^0 + (-1) * 3^1 + (-1) * 3^2 + (-1) * 3^3 + 1 * 3^4 = 42`;
+- `byte < 0`: use `byte + 243` as an index for the `LUT` e.g. if the `byte = -42` then `LUT[201] = [ 0, 1, 1, 1, -1 ]` and `0 * 3^0 + 1 * 3^1 + 1 * 3^2 + 1 * 3^3 + (-1) * 3^4 = -42`;
+
+Even though this RFC won't cover every little implementation detail, it can be noted that the `LUT` could cover all `256` possible values of a `byte` to avoid invalid accesses. To the missing `byte` values we can simply assign some default mapping to `5` `trits`, eg. `[ 0, 0, 0, 0, 0 ]`. This way runtime checks can be avoided as branching on each iteration of a loop can degrade performance in case of false branch prediction.
+
+Additionally, there could be two variants of the decoding function:
+- checked & slow;
+- unchecked & fast;
```
wusyong, line 92: I think unchecked is probably a `unsafe` function or method. Is there anything we need to consider if calling unchecked? We can also add in documentation when implementing.

text/0013-trits-t5b1/0013-trits-t5b1.md
```diff
@@ -0,0 +1,324 @@
...
+
+The overhead of the `13` additional values is only +~5% to the size of `LUT`.
```
wusyong, line 94: We don't really need these values in LUT. Match patter can specify them e.g. `_ => LUT[0]`

thibault-martinez: The goal of these values in LUT is *specifically* to avoid pattern matching which would slow down conversion.

text/0013-trits-t5b1/0013-trits-t5b1.md
thibault-martinez: I'd like, whenever possible, to not necessarily tie RFCs to Rust.

text/0013-trits-t5b1/0013-trits-t5b1.md
thibault-martinez: Unchecked should only be called on trusted sources i.e. things we converted ourselves and from a checked function. One shouldn't call an unchecked function directly on network data.

wusyong (approved):

PR #14, opened 2019-09-25
RFC: `trits-t1b1` encoding

Alex6323:
[Rendered](https://github.com/Alex6323/bee-rfcs/blob/rfc-t1b1/text/0000-trits-t1b1/0000-trits-t1b1.md)

### DIFF
--- a/text/0000-trits-t1b1/0000-trits-t1b1.md
+++ b/text/0000-trits-t1b1/0000-trits-t1b1.md
@@ -1,28 +1,98 @@
 + Feature name: `trits-t1b1`
-+ Start date: 2019-09-25)
++ Start date: 2019-09-25
 + RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
 + Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
 
 # Summary
 
-This RFC proposes an encoding that stores a single trit (`-1`, `0`, `1`) in a signed byte `i8`.
+This RFC proposes an encoding that stores sequences of balanced trits (`-1`, `0`, `1`) in a `Vec` of signed bytes `i8`. It is one of the encodings used in the IOTA ecosystem, which are mentioned in RFC #10. It is the most memory inefficient encoding possible to store or transfer trits, because only 3 out of 256 possible states in a byte are actually used. Nevertheless this sometimes referred to as "raw trits" encoding is the representation expected as input by the trinary `Curl` hash function, and as well the representation of its output. 
 
 # Motivation
 
-WIP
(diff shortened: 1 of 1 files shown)


text/0000-trits-t1b1/0000-trits-t1b1.md
```diff
@@ -0,0 +1,98 @@
...
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+This RFC proposes an encoding that stores sequences of balanced trits (`-1`, `0`, `1`) in a `Vec` of signed bytes `i8`. It is one of the encodings used in the IOTA ecosystem, which are mentioned in RFC #10. It is the most memory inefficient encoding possible to store or transfer trits, because only 3 out of 256 possible states in a byte are actually used. Nevertheless this sometimes referred to as "raw trits" encoding is the representation expected as input by the trinary `Curl` hash function, and as well the representation of its output. 
+
+# Motivation
+
+This encoding of sequences of balanced trits is necessary for the `Curl` hash function and therefore still necessary to allow for the Bee framework to interoperate with the current IOTA mainnet.
```
semenov-vladyslav, line 12: `Curl` can be implemented using a more efficient encoding (64 trits per two 64-bit words). This way memory footprint can be reduced by a factor of 4. When pointer arithmetic is used to access words of `Curl` state (instead of `LUT`) then the memory footprint reduces additionally be a factor of 4/3. All in all, optimizing of `Curl` is worth investigating.

So, it's not a very convincing argument that "balanced trits is necessary for the `Curl`". Maybe we should talk about `Curl`-compliant encoding?

`t1b1` however is the simplest encoding that could be very useful for tests, to compare invariants when other encodings are used to perform the same trits transformation.

thibault-martinez: Not if we actually do a difference between `Curl` and `PCurl`. Per se `Curl` is the one operating on `T1B1` while `PCurl` accepts different encodings. It is briefly mentioned in the base RFC here https://github.com/iotaledger/bee-rfcs/pull/10/files#diff-b075c83057cb59013a66d6ee87aa3986R31.

Alex6323: yea, I know of BCT-Curl. There is already a Rust port of it btw. But can we use it as a full replacement for Standard-Curl? Or putting this question differently: Is there really no other usecase for `T1B1` other than tests, debugging, ....

thibault-martinez: I think for single hashing `Curl` is faster than `PCurl` because of all the overhead it has. Also, almost all prototyping is done with `T1B1`.

PR #15, opened 2019-09-26
RFC: Bee-rfcs README

jakubcech:
[Rendered](https://github.com/iotaledger/bee-rfcs/blob/7e7c5ef23a57f6f02382becbae52a7f63337ba17/text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md) (is this how it's done? 😃) 

An RFC proposing we outline a consistent way of writing RFC motivations.

### DIFF
--- a/text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
+++ b/text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
@@ -5,7 +5,7 @@
 
 # Summary
 
-This RFC proposes a change to the RFC process itself. Particularly the `Motivation` section of the RFC template to put more emphasis on the need to write clear requirements and how to write them.
+This RFC proposes extending the "Motivation" section to put more emphasis on the need to write clear requirements and suggest how to write them.
 
 # Motivation
 
@@ -25,7 +25,9 @@ With the following:
 Why are we doing this? What use cases does it support? What is the expected outcome?
 
 1. Write a summary of the motivation.
-2. List all the specific use cases that your proposal is trying to address. Write them from the perspective of the person who will be using the software. We recommend using the "Job story" format, which provides context, motivation, and the desired outcome.
+2. List all the specific use cases that your proposal is trying to address. 
+3. Where applicable, write from the perspective of the person who will be using the software, for example using the "Job story" format:
+
 
 When _____ , I want to _____ , so I can _____ .
 
--- a/text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
+++ b/text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
@@ -1,0 +1,58 @@
++ Feature name: `BEE RFCs readme motivation change`
++ Start date: 2019-09-26
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+This RFC proposes a change to the RFC process itself. Particularly the `Motivation` section of the RFC template to put more emphasis on the need to write clear requirements and how to write them.
+
+# Motivation
+
++ A more consistent and better means of capturing the motivations driving the changes proposed in RFCs across all projects.
++ Ensuring the changes proposed in RFCs are thought through from the user and business perspective.
++ Stakeholders should be involved in the RFC process as early as a new RFC is created. A well defined motivation ensures that ideas can be easily validated before the design is scrutinized.
+
+# Detailed design
+
+Replace the current motivation in the template:
+```
+Why are we doing this? What use cases does it support? What is the expected outcome?
+```
+
+With the following:
+```
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+1. Write a summary of the motivation.
+2. List all the specific use cases that your proposal is trying to address. Write them from the perspective of the person who will be using the software. We recommend using the "Job story" format, which provides context, motivation, and the desired outcome.
+
+When _____ , I want to _____ , so I can _____ .
+
+Example 1:
+When I query a node for a list of transactions, I want to be able to sort them by date, so I can work with the most relevant ones.
+
+Example 2:
+When I configure a node, I want to be able to control how much transaction history the node stores, so I can make sure I only store the data I need without incurring additional operational costs.
+
+```
+
+The suggested format is an evolution on a user story format. It may not apply to all RFCs, but I'd encourage everyone to try and think of any functionality with this framing.
+
+# Drawbacks
+
+- The motivation part of RFCs will take longer to write up.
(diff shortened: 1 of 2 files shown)


text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
```diff
@@ -0,0 +1,58 @@
...
+# Rationale and alternatives
+
+- The job story approach forces thinkink from the perspective of the person who will be using the product.
+- Standard requirement format was considered, but it allows for too much vagueness.
+- Acceptance criteria format was considered, but that is mainly suitable for products with UI or very UX heavy products. 
+- What is the impact of not doing this?: Not doing this I'm afraid will lead to incosistent motivation specifications and disallow us from quickly validating individual RFCs from the motivation perspective.
+
+# Unresolved questions
+
+- Do we want to enforce this in a format like this? Or should we use a more loose format?
```
wusyong, line 60: I think just make this loose format. Let draftsman has some clues on how to write it at least.

jakubcech: I'm all for suggesting and not enforcing at all costs.

text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
```diff
@@ -0,0 +1,58 @@
...
+```
+Why are we doing this? What use cases does it support? What is the expected outcome?
+```
+
+With the following:
+```
+Why are we doing this? What use cases does it support? What is the expected outcome?
+
+1. Write a summary of the motivation.
+2. List all the specific use cases that your proposal is trying to address. Write them from the perspective of the person who will be using the software. We recommend using the "Job story" format, which provides context, motivation, and the desired outcome.
```
SuperFluffy: Replace this with:

2. List all the specific use cases that your proposal is trying to address.
3. Where applicable, write from the perspective of the person who will be using the software, for example using the "Job story" format:

text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
```diff
@@ -0,0 +1,58 @@
...
+Example 2:
+When I configure a node, I want to be able to control how much transaction history the node stores, so I can make sure I only store the data I need without incurring additional operational costs.
+
+```
+
+The suggested format is an evolution on a user story format. It may not apply to all RFCs, but I'd encourage everyone to try and think of any functionality with this framing.
+
+# Drawbacks
+
+- The motivation part of RFCs will take longer to write up.
```
SuperFluffy, line 46: I don't think that's the case. The suggested changes might actually help to make the process clearer for the writer.

text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
```diff
@@ -0,0 +1,58 @@
...
+- Larger RFCs may tackle large amounts of 'jobs'. In which case it might be beneficial to either break down the RFC, or list only higher level motivations.
```
SuperFluffy, line 47: This also seems orthogonal to the proposal. If RFCs touch very many different concerns, they should be broken up, if possible.

text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
```diff
@@ -0,0 +1,58 @@
...
+- For the current implementation of 'known components', this may seem redundant. If the RFCs motivation is really clear, however, it should not be an issue to write such a motivation.
```
SuperFluffy, line 48: Motivating a change is always recommended because what seems obvious to a long-time user of IOTA might not be clear to a newbie. If the particular proposed change doesn't fit the motivation steps as laid out here, people are always free to not adhere to them.

text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
```diff
@@ -0,0 +1,58 @@
...
+- If it's an internal component, it may be difficult to think from the user perspective. However, even internal components need to tie into a specific scenario that they allow.
```
SuperFluffy, line 49: Hopefully there is always an advantage to the end user, for example in terms of performance, flexibility, or stability, so that part can be motivated.

I agree that it might be difficult to come up with a “user story”. So we shouldn't enforce this too much.

text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
SuperFluffy: I have changed the wording of the “job story” format somewhat. This steers proposal writers into the right direction, but leaves it flexible enough that people will not be discouraged.

Since we want to evaluate every RFC anyways, we can always point out that a “user story” motivation would have been possible (if that's the case), and request the RFC to be amended thus.

text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
```diff
@@ -0,0 +1,58 @@
++ Feature name: `BEE RFCs readme motivation change`
++ Start date: 2019-09-26
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+This RFC proposes a change to the RFC process itself. Particularly the `Motivation` section of the RFC template to put more emphasis on the need to write clear requirements and how to write them.
```
SuperFluffy: This does not seem to change the RFC process itself. This RFC seems more like a clarification and suggestion as to how to write RFCs.

SuperFluffy (changes requested):
I like the proposed changes.

Other than written in the RFC, I note that this RFC does not seem to change the RFC process itself. The proposed changes to the template seem to rather clarify the intent we are looking for in a new RFC.

As such, I think we could safely close this RFC and merge the changes without the RFC process.


@jakubcech I would ask you to edit `0000-template.md` with your proposed changes (and possibly my suggestions) and commit them to this PR.

jakubcech:
Committed a change to the template itself.
Made changes to the Motivation steps as proposed.

text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md
jakubcech: Done

SuperFluffy (approved):

SuperFluffy:
I approve of this proposal, but I don't think it warrants its own RFC because I don't see a fundamental change in the actual RFC procedure.

I personally vote for closing this RFC, submitting the modified template as a new PR and putting the argument into the PR text.

Alternatively, just modify this PR and remove `text/0000-bee-rfcs-readme/0000-bee-rfcs-readme.md`, adding its text to the top comment.

jakubcech:
Created a new PR: 
https://github.com/iotaledger/bee-rfcs/pull/17

PR #18, opened 2019-10-02
RFC: Bundle type

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/bee-bundle/text/0018-bee-bundle/0018-bee-bundle.md)

### DIFF
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -60,7 +60,9 @@ impl Bundle {
 
 ## Algorithms
 
-<!-- TODO -->
+In this section, we describe the algorithms needed to build a `Bundle`. The lifecycle of a `BundleBuilder` depends on if it's being used in client side or server side:
+- client side: `finalise` -> [`sign` ->] `validate` -> `build`
+- server side: `addTransaction`/`addTransactionBuilder` -> `validate` -> `build`
 
 ### Bundle hash generation
 
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -36,13 +36,17 @@ Transactions are final and bundles, essentially being arrays of transactions, ar
 
 ### Bundle
 
-A bundle can simply be represented as an array of transactions. As bundles are final, they shouldn't be modifiable outside of the scope of the bundle module.
+As bundles are final, they shouldn't be modifiable outside of the scope of the bundle module.
+
+A bundle can simply be represented as an array of transactions:
 
 ```rust
 struct Bundle {
     transactions: Vec<Transaction>
 }
 ```
+
+And its implementation should only allow to retrieve transactions:
 
 ```rust
 impl Bundle {    
@@ -51,8 +55,6 @@ impl Bundle {
     }
 }
 ```
-
-<!-- TODO -->
 
 ### BundleBuilder
 
@@ -139,7 +141,7 @@ sign(bundle, seed, inputs)
 | | | | input = inputs[transaction.address]
 | | | | fragments = sign(seed, input.index, input.security, transaction.bundle)
 | | | | for fragment in fragments
-| | | | | bundle[current_index].signature = fragment
+| | | | | bundle.at(current_index).signature = fragment
 | | | | | current_index = current_index + 1
 | | else
 | | | current_index = current_index + 1
@@ -222,3 +224,4 @@ validate(bundle):
 <!-- TODO -->
 
 - Should this RFC expands a bit more on the M-Bug ? Or give a link ?
+- Should `Bundle` provide a `.transactions` or a `.at` method ?
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -64,7 +64,7 @@ In this section, we describe the algorithms needed to build a `Bundle`. The life
 - client side: `finalise` -> [`sign` ->] `validate` -> `build`
 - server side: `addTransaction`/`addTransactionBuilder` -> `validate` -> `build`
 
-### Bundle hash generation
+### Hash
 
 *Client side and server side operation.*
 
@@ -86,7 +86,7 @@ The bundle hash is generated with a sponge by iterating through the bundle, from
 Pseudocode:
 
 ```
-bundleHash(bundle)
+hash(bundle)
 | sponge = Sponge(HASH_FUNCTION)
 |
 | for transaction in bundle
@@ -97,7 +97,7 @@ bundleHash(bundle)
 
 *In the current mainnet, the hash function of the sponge used to generate bundle hashes is Kerl.*
 
-### Bundle finalisation
+### Finalise
 
 *Client side operation.*
 
@@ -106,7 +106,7 @@ Finalising a bundle means computing the bundle hash, verifying that it matches t
 Pseudocode:
 
 ```
-bundleFinalise(bundle)
+finalise(bundle)
 | hash = bundleHash(bundle)
 |
 | while hash.normalise().find('M')
@@ -119,7 +119,7 @@ bundleFinalise(bundle)
 
 *Security requirement: due to the implementation of the signature process, the normalised bundle hash can't contain a `M` or `13` because it could expose a significant part of the private key, weakening the signature. The bundle hash is then repetitively generated with a slight modification until its normalisation doesn't contain a `M`.*
 
-### Bundle validation
+### Validate
 
 *Server side operation.*
 
@@ -140,7 +140,7 @@ For a bundle to be considered valid, the following assertions must be true:
 Pseudocode:
 
 ```
-bundleValidate(bundle):
+validate(bundle):
 | value = 0
 | current_index = 0
 |
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -253,11 +253,11 @@ validate(bundle):
 
 # Rationale and alternatives
 
-<!-- TODO -->
+- A `Bundle` is a fundamental component of the IOTA protocol and must be implemented;
+- There is no more intuitive and simple way to implement a `Bundle` than the one proposed;
+- Since bundles are final, `BundleBuilder` is mandatory;
 
 # Unresolved questions
-
-<!-- TODO -->
 
 - Should this RFC expands a bit more on the M-Bug ? Or give a link ?
 - Should `Bundle` provide a `.transactions` or a `.at` method ?
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -5,6 +5,12 @@
 
 # Summary
 
+Useful links:
+- [Bundles and transactions
+](https://docs.iota.org/docs/dev-essentials/0.1/concepts/bundles-and-transactions)
+- [What is a bundle?
+](https://docs.iota.org/docs/getting-started/0.1/introduction/what-is-a-bundle)
+
 <!-- TODO -->
 
 # Motivation
@@ -12,6 +18,39 @@
 <!-- TODO -->
 
 # Detailed design
+
+## Bundle hash generation
+
+The hash of a bundle is based on the bundle essence of each of its transactions. The bundle essence of a transaction is composed of the following fields.
+
+| Name          | Size      |
+| ------------- | --------- |
+| address       | 243 trits |
+| value         | 81 trits  |
+| obsolete_tag  | 81 trits  |
+| timestamp     | 27 trits  |
+| current_index | 27 trits  |
+| last_index    | 27 trits  |
+
+The bundle hash is generated with a sponge by iterating through the bundle, from `0` to `last_index`, absorbing the bundle essence of each transaction and eventually squeezing the bundle hash from the sponge.
+
+Pseudocode:
+
+```
+bundleHash(bundle)
+| sponge = Sponge(HASH_FUNCTION)
+| for transaction in bundle
+| | sponge.absorb(transaction.essence())
+| return sponge.squeeze()
+```
+
+*In the current mainnet, the hash function of the sponge used to generate bundle hashes is Kerl.*
+
+## Bundle finalisation
+
+<!-- TODO -->
+
+## Bundle validation
 
 <!-- TODO -->
 
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -119,6 +119,28 @@ finalise(bundle)
 
 *Security requirement: due to the implementation of the signature process, the normalised bundle hash can't contain a `M` or `13` because it could expose a significant part of the private key, weakening the signature. The bundle hash is then repetitively generated with a slight modification until its normalisation doesn't contain a `M`.*
 
+### Sign
+
+Pseudocode:
+
+```
+sign(bundle, seed, inputs)
+| current_index = 0
+|
+| for transaction in bundle
+| | if transaction.value < 0
+| | | if transaction.current_index >= current_index
+| | | | input = inputs[transaction.address]
+| | | | fragments = sign(seed, input.index, input.security, transaction.bundle)
+| | | | for fragment in fragments
+| | | | | bundle[current_index].signature = fragment
+| | | | | current_index = current_index + 1
+| | else
+| | | current_index = current_index + 1
+```
+
+<!-- TODO -->
+
 ### Validate
 
 *Server side operation.*
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -29,6 +29,8 @@ In this section, we present the main algorithms needed to process bundles, serve
 
 ## Algorithms
 
+<!-- TODO -->
+
 ### Bundle hash generation
 
 The hash of a bundle is based on the bundle essence of each of its transactions. The bundle essence of a transaction is composed of the following fields.
@@ -132,6 +134,8 @@ bundleValidate(bundle):
 | return BUNDLE_VALID
 ```
 
+<!-- TODO -->
+
 ## Bundle and BundleBuilder
 
 <!-- TODO -->
@@ -156,5 +160,4 @@ bundleValidate(bundle):
 
 <!-- TODO -->
 
-- Should this RFC include explanation and/or pseudocode for bundle hash normalisation ?
 - Should this RFC expands a bit more on the M-Bug ? Or give a link ?
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -189,7 +189,7 @@ sign(bundle, seed, inputs)
 
 *Client side operation.*
 
-Proof of Work (PoW) allows your transactions to be accepted by the network. On the IOTA network, PoW is only a rate control mechanism. Doing PoW on a bundle means doing PoW on each of its transactions and set trunks and branch accordingly. After PoW, a bundle is ready to be sent to the network.
+Proof of Work (PoW) allows your transactions to be accepted by the network. On the IOTA network, PoW is only a rate control mechanism. Doing PoW on a bundle means doing PoW on each of its transactions and setting trunks and branch accordingly. After PoW, a bundle is ready to be sent to the network.
 
 Pseudocode:
 
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -11,7 +11,7 @@ This RFC is based on [`transaction-module`]().
 The smallest communication unit in the IOTA protocol is the transaction. Everything, including payment settlements
 and/or plain data, is propagated through the IOTA network in transactions.
 
-A transaction is `2673` trytes and the part available to the user is `2187` trytes. This part holds a signature in case
+A transaction is `8019` trits and the part available to the user is `6561` trits. This part holds a signature in case
 of a payment settlement and plain data otherwise. Since it has a limited size, a user often needs more than one
 transaction to fulfil their operation, for example signatures with security level `2` or `3` don't fit in a single
 transaction and user-provided data may exceed the allowance so they need to be fragmented across multiple transactions.
--- a/text/0018-bee-bundle/0018-bee-bundle.md
+++ b/text/0018-bee-bundle/0018-bee-bundle.md
@@ -10,7 +10,7 @@ This RFC is based on [`transaction-module`]().
 
 The smallest communication unit in the IOTA protocol is the transaction. Everything, including payment settlements and/or plain data, is propagated through the IOTA network in transactions.
 
-A transaction is `2673` trytes and the part available to the user is `2187` trytes. This part holds a signature in case of a payment settlement and plain data otherwise. Since it has a limited size, a user often needs more than one transaction to fulfil his operation, for example signatures with security level `2` or `3` don't fit in a single transaction and user-provided data may exceed the allowance so they need to be fragmented across multiple transactions. Moreover, a transaction may not make sense on its own, for example an input/output transaction alone would change the total amount of the ledger so it has to be paired with another complementary input/output.
+A transaction is `2673` trytes and the part available to the user is `2187` trytes. This part holds a signature in case of a payment settlement and plain data otherwise. Since it has a limited size, a user often needs more than one transaction to fulfil their operation, for example signatures with security level `2` or `3` don't fit in a single transaction and user-provided data may exceed the allowance so they need to be fragmented across multiple transactions. Moreover, a value transaction doesn't make sense on its own because it would change the total amount of the ledger so it has to be paired with other complementary transactions that together will balance the total value to zero.
 
 For these reasons, transactions have to be processed as a whole, in groups called bundles. A bundle is an atomic operation in the sense that either all or none of its transactions are accepted by the network. Even single transactions are propagated through the network within a bundle making it the only confirmable communication unit of the IOTA protocol.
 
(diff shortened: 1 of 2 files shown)


text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,171 @@
++ Feature name: `bee-bundle`
++ Start date: 2019-10-02
++ RFC PR: [iotaledger/bee-rfcs#18](https://github.com/iotaledger/bee-rfcs/pull/18)
++ Bee issue: [iotaledger/bee#63](https://github.com/iotaledger/bee/issues/63)
+
+# Summary
+
+The smallest communication unit in the IOTA protocol is the transaction. Everything, including payment settlements and/or plain data, is propagated through the IOTA network in transactions.
+
+A transaction is `2673` trytes and the part available to the user is `2187` trytes. This part holds a signature in case of a payment settlement and plain data otherwise. Since it has a limited size, a user often needs more than one transaction to fulfil his operation, for example signatures with security level `2` or `3` don't fit in a single transaction and user-provided data may exceed the allowance so they need to be fragmented across multiple transactions. Moreover, a transaction may not make sense on its own, for example an input/output transaction alone would change the total amount of the ledger so it has to be paired with another complementary input/output.
```
tsvisabo: "his" => "his/her"
"a transaction may not make sense on its own" => "a value transaction will not make sense on its own because .."

tsvisabo: "..complementary input/output"  + "that together will balance the balance total value to zero"

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,171 @@
...
+| Name          | Size      |
+| ------------- | --------- |
+| address       | 243 trits |
+| value         | 81 trits  |
+| obsolete_tag  | 81 trits  |
+| timestamp     | 27 trits  |
+| current_index | 27 trits  |
+| last_index    | 27 trits  |
+
+The bundle hash is generated with a sponge by iterating through the bundle, from `0` to `last_index`, absorbing the bundle essence of each transaction and eventually squeezing the bundle hash from the sponge.
```
tsvisabo: sponge function

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,171 @@
...
+
+Validating a bundle means checking the syntactic and semantic integrity of a bundle as a whole and of its constituent transactions. As bundles are atomic transfers, either all or none of the transactions will be accepted by the network. After validation, transactions of a bundle are candidates to be included to the ledger.
+
+For a bundle to be considered valid, the following assertions must be true:
+
+- bundle has announced size;
+- transactions share the same bundle hash;
+- transactions absolute value doesn't exceed total IOTA supply;
+- bundle absolute sum never exceeds total IOTA supply;
+- transactions appear in the announced order;
```
tsvisabo: what does it mean `announced`? index?
what does it mean `appear`?

tsvisabo (changes requested):

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Thanks, I'll go with they/their which is usually the recommended gender neutral form.

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Technically, I'm speaking of a sponge object as seen in the pseudocode, so I think it's fine.

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Rephrased

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
++ Bee issue: [iotaledger/bee#63](https://github.com/iotaledger/bee/issues/63)
+
+# Summary
+
+This RFC is based on [`transaction-module`]().
+<!-- TODO FIx link to RFC#3 -->
+
+The smallest communication unit in the IOTA protocol is the transaction. Everything, including payment settlements and/or plain data, is propagated through the IOTA network in transactions.
+
+A transaction is `2673` trytes and the part available to the user is `2187` trytes. This part holds a signature in case of a payment settlement and plain data otherwise. Since it has a limited size, a user often needs more than one transaction to fulfil their operation, for example signatures with security level `2` or `3` don't fit in a single transaction and user-provided data may exceed the allowance so they need to be fragmented across multiple transactions. Moreover, a value transaction doesn't make sense on its own because it would change the total amount of the ledger so it has to be paired with other complementary transactions that together will balance the total value to zero.
```
SuperFluffy: Let's express this in `trits`. Further down, the RFC talks about `trits` instead of `trytes`. I think `trits` is the more natural/less confusing unit to talk about. `trit` already puts some mental overhead on the user, because it's very unusual, and `tryte` being defined as a multiple of `trit` even more so.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+For these reasons, transactions have to be processed as a whole, in groups called bundles. A bundle is an atomic operation in the sense that either all or none of its transactions are accepted by the network. Even single transactions are propagated through the network within a bundle making it the only confirmable communication unit of the IOTA protocol.
+
+This RFC proposes ways to create and manipulate a bundle and describe the associated algorithms.
+
+Useful links:
```
SuperFluffy, line 25: Maybe we should also put a link to trits here, or at least quickly define them?

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+### Bundle
+
+As bundles are final, they shouldn't be modifiable outside of the scope of the bundle module.
+
+A bundle can simply be represented as an array of transactions:
+
+```rust
+struct Bundle {
+    transactions: Vec<Transaction>
```
SuperFluffy: Is there a natural order to `transactions`? Should they be included one transaction at a time? Is there a chance that transactions arrive out of order? Can we use a `HashMap` instead of `Transaction` here?

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+}
+```
+
+And its implementation should only allow to retrieve transactions:
+
+```rust
+impl Bundle {    
+    pub fn transactions(&self) -> &Vec<Transaction> {
+        &self.transactions
```
SuperFluffy, line 77: Should we hide this as an implementation detail and instead provide a `struct Transactions(Vec<Transaction>)` newtype?

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+}
+```
+
+### BundleBuilder
+
+The `BundleBuilder` offers a simple and convenient way to build bundles:
+
+```rust
+pub struct BundleBuilder {
+    transactions: Vec<TransactionBuilder>
```
SuperFluffy, line 88: Is there a semantic difference between an empty vector and a vector that hasn't been set? If yes, I would set this to `Option<Vec<TransactionBuilder>>` (or to `Option<TransactionBuilders>`, if we want to hide the details).

Please also note my proposal for constructing a `Transaction` via a `UnprovenTransaction` at https://github.com/rufsam/bee-rfcs/pull/1

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+```
+
+*We do not list parameters and/or return values as they are implementation details.*
+
+## Algorithms
+
+In this section, we describe the algorithms needed to build a `Bundle`. The lifecycle of a `BundleBuilder` depends on if it's being used in client side or server side:
+- client side: `finalise` -> [`sign` ->] -> [`pow` ->] -> `validate` -> `build`
```
SuperFluffy: Corrected"
`finalise -> [sign ->] [pow ->] validate -> build`

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+- server side: `addTransaction`/`addTransactionBuilder` -> `validate` -> `build`
```
SuperFluffy: Rust convention is `small_snake_case`, `add_transaction` and `add_transaction_builder`.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+*`sign` is optional because data transactions don't have to be signed. `pow` is optional because one can use remote pow instead.*
+
+### Hash
+
+*Client side and server side operation.*
+
+A bundle hash ties different transactions together. By having this common hash in their `bundle` field, it makes it clear that these transactions should be processed as a whole.
```
SuperFluffy: If I read this correctly, this is a hard requirement on the design of `Transaction`. If that's the case, we might need to combine the RFCs after all.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+The hash of a bundle is derived from the bundle essence of each of its transactions. The bundle essence of a transaction is composed of the following fields.
```
SuperFluffy: We should make it explicit that `bundle essence` refers to a subset of fields comprising the transaction.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+| Name          | Size      |
+| ------------- | --------- |
+| address       | 243 trits |
```
SuperFluffy, line 144: For consistency, we should decide whether we want to use `trits` or `trytes` here (see also the review comment of the summary section).

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+Pseudocode:
+
+```
+hash(bundle)
+| sponge = Sponge(HASH_FUNCTION)
+|
+| for transaction in bundle
+| | sponge.absorb(transaction.essence())
+|
+| return sponge.squeeze()
```
SuperFluffy, line 163: I'd prefer to not have pipes in here, as the whitespace is sufficient to distinguish blocks.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+|
+| while hash.normalise().find('M')
+| | bundle.at(0).obsolete_tag++
+| | hash = bundleHash(bundle)
+|
+| for transaction in bundle
+| | transaction.setBundleHash(hash)
+```
+
+*Security requirement: due to the implementation of the signature process, the normalised bundle hash can't contain a `M` or `13` because it could expose a significant part of the private key, weakening the signature. The bundle hash is then repetitively generated with a slight modification until its normalisation doesn't contain a `M`.*
```
SuperFluffy: This is a sufficiently weird caveat that we should definitely further explain it or point to the revelant documents explaining it.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+```rust
+pub struct BundleBuilder {
+    transactions: Vec<TransactionBuilder>
+}
+```
+
+```rust
+impl BundleBuilder {    
+    pub fn hash(&self) {
```
SuperFluffy: In cases where the verb isn't clear, I would opt to prepend a method with `calculate_`, making this `calculate_hash`.

Also for expensive operations it signifies that some work will be done.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+    pub fn finalise(&self) {
+      unimplemented!()
+    }
+
+    pub fn sign(&self) {
+      unimplemented!()
+    }
+
+    pub fn pow(&self) {
```
SuperFluffy: As for `hash`, I would call this `calculate_proof_of_work`. We might consider passing a function `F: Fn(&[u8]) -> Nonce` here to allow for different PoW algorithms.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+*Client side operation.*
+
+Finalising a bundle means computing the bundle hash, verifying that it matches the security requirement and setting it to all the transactions of the bundle. After finalisation, transactions of a bundle are ready to be safely attached to the tangle.
+
+Pseudocode:
+
+```
+finalise(bundle)
+| hash = bundleHash(bundle)
```
SuperFluffy, line 178: Same as above, for these short blocks I think the extra `|` produce too much noise.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+- order of transactions in the bundle is the same as announced by `current_index` and `last_index`;
+- value transactions have an address ending in `0` i.e. has been generated with Kerl;
+- bundle inputs and outputs are balanced i.e. the bundle sum equals `0`;
+- announced bundle hash matches the computed bundle hash;
+- for spending transactions, the signature is valid;
+
+Pseudocode:
+
+```
+validate(bundle):
```
SuperFluffy, line 270: Because a `Bundle` should be a valid object, the validation step should be part of the `build` method.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+# Rationale and alternatives
+
+- A `Bundle` is a fundamental component of the IOTA protocol and must be implemented;
+- There is no more intuitive and simple way to implement a `Bundle` than the one proposed;
+- Since bundles are final, `BundleBuilder` is mandatory;
+
+# Unresolved questions
+
+- Should this RFC expands a bit more on the M-Bug ? Or give a link ?
```
SuperFluffy, line 317: Yes.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+- Should `Bundle` provide a `.transactions` or a `.at` method ?
```
SuperFluffy, line 318: Is there a natural order to `Bundle`? Can `Transactions` arrive out of order?

The `.at` method is basically indexing into the `Bundle`, getting a specific transaction? I think this makes sense, if there is a natural order.

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Will do it for the sake of consistency. But when talking about user-facing sizes we usually prefer trytes over trits because of the nice readable mapping. And trits are usually used in crypto contexts.

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Sure

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: There is a natural order. Each transaction has a `current_index` and a `last_index`. So transactions in a bundle should be ordered from `0` to `last_index`.

> Is there a chance that transactions arrive out of order

Yes definitely, but we don't really care because:
- client-side: we add transaction one at a time
- server-side: we process bundle only when we know all the transactions are present

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Good idea

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: I don't think there is a difference, in both case it'll fail to build a bundle

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Whoops, thanks

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Thanks

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Algorithms in papers usually use bars but alright, whitespaces

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Yeah I know :/
But at the same time this should and will be expanded in the actual Signing Scheme RFC
The perfect way would be later to link to the RFC

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Sure

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: `build` will definitely call `validate`. But I assumed `validate` is a function sufficiently interesting be able to be called on its own. I can imagine situation when we want to validate but doesn't care of the resulting bundle.

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: Yes there is, `.at()` it is then.

text/0018-bee-bundle/0018-bee-bundle.md
samuel-rufi: > value transaction doesn't make sense on its own

maybe call it` payment settlement` instead `value transaction`. We should distinguish `transaction`, the atomic unit, from a settlement which requires multiple transactions. The term "transaction" should not be used as synonym for `bundle`, else it confuses only.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+hash(bundle)
+| sponge = Sponge(HASH_FUNCTION)
+|
+| for transaction in bundle
+| | sponge.absorb(transaction.essence())
+|
+| return sponge.squeeze()
+```
+
+*In the current mainnet, the hash function of the sponge used to generate bundle hashes is Kerl.*
```
samuel-rufi: Not sure if we should write "in the current mainnet". I mean this bundle proposal is directly for the mainnet. We shouldn't need to refer it.

text/0018-bee-bundle/0018-bee-bundle.md
samuel-rufi: Agree, and why it's not possible to use all transaction fields. Moreover a reader might ask why `bundle essence` and not simply `bundle`.

text/0018-bee-bundle/0018-bee-bundle.md
wusyong: Yes, we might receive the transaction out of order. I think stay in vector makes sense. Otherwise, we shouldn't just use `HashMap` and should build our own linked list furthermore.

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,281 @@
...
+
+*We do not list parameters and/or return values as they are implementation details.*
+
+## Algorithms
+
+In this section, we describe the algorithms needed to build a `Bundle`. The lifecycle of a `BundleBuilder` depends on if it's being used in client side or server side:
+- client side: `finalise` -> [`sign` ->] -> [`pow` ->] -> `validate` -> `build`
+- server side: `addTransaction`/`addTransactionBuilder` -> `validate` -> `build`
+
+*`sign` is optional because data transactions don't have to be signed. `pow` is optional because one can use remote pow instead.*
```
wusyong: As we introduce intermediate type, I think pow validation will be mandatory before built? Besides, even it's remote pow, client still needs to wait for returned nonce no?

text/0018-bee-bundle/0018-bee-bundle.md
wusyong: There are also other ways to form a bundle like `Entry`. While I'm not sure we are also going this way, I would suggest these should be separate RFC.

text/0018-bee-bundle/0018-bee-bundle.md
wusyong: I would recommend stay in trytes and maybe just mention it's 3 trits 1 trytes. Many documentations and other libraries across IOTA ecosystem are still in form of trytes when it comes to build transaction and bundles.

text/0018-bee-bundle/0018-bee-bundle.md
wusyong: Might also want some trait bound or trait object. But I guess this depends on design of POW module. Let's just stay this for now. Further discussion should bring to #19.

wusyong: Btw, I think we should pass `mwm` too.

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: @rufsam But here `value transaction` is used on purpose to indeed make the point that we need bundles.

samuel-rufi: @thibault-martinez okey understand what you mean. however, in general, we shouldn't use the term `value transaction` for transfering value, and maybe make this clear in the RFC.

text/0018-bee-bundle/0018-bee-bundle.md
zesterer: Alternatively, you could return either `&[Transaction]` or `impl FixedSizeIterator<Item=&Transaction>`

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,308 @@
...
+
+# Summary
+
+This RFC is based on [`transaction-module`]().
+<!-- TODO FIx link to RFC#3 -->
+
+The smallest communication unit in the IOTA protocol is the transaction. Everything, including payment settlements
+and/or plain data, is propagated through the IOTA network in transactions.
+
+A transaction is `2673` trytes and the part available to the user is `2187` trytes. This part holds a signature in case
```
Alex6323: * maybe we should also mention the byte equivalent? ~1605 bytes (in the most compact encoding)
* "the part available to the user" -> should we just use the common term "payload" to describe that?

thibault-martinez: Janis wants trits, you want bytes, Yu-Wei wants trytes... :trollface: 
Thanks, I like payload.

Alex6323: regarding the byte equivalent: I had my IOTA-newcomer glasses on :smile:

text/0018-bee-bundle/0018-bee-bundle.md
```diff
@@ -0,0 +1,308 @@
...
+of a payment settlement and plain data otherwise. Since it has a limited size, a user often needs more than one
+transaction to fulfil their operation, for example signatures with security level `2` or `3` don't fit in a single
+transaction and user-provided data may exceed the allowance so they need to be fragmented across multiple transactions.
```
Alex6323, line 14: This is gonna be a bit controversial, but I feel this is the only point in time where we can change a fundamental term:

Isn't essentially the `bundle` what should be called a `transaction` according to the database definition of a transaction where a transaction is an *atomic* operation. But in IOTA's case it is the bundle that is atomic, not the transaction, right?
Atomicity: A transaction must be fully complete, saved (committed) or completely undone (rolled back). A sale in a retail store database illustrates a scenario which explains atomicity, e.g., the sale consists of an inventory reduction and a record of incoming cash. Both either happen together or do not happen - it's all or nothing.

Likewise either all the transactions in a bundle get confirmed, or none of them.

The reasoning for bringing this up is again looking on the proposal from the outside as someone who might not be familiar with IOTA terms, and making its terminology most accessible to newcomers. 

It doesn't sound as sexy but I would propose the terms `Transaction` (former bundle) and `Transaction Fragment` (former transaction), while the `Signature Or Message Fragment` is renamed to `Payload`.

If I have made an essential mistake here, then please correct me. Wouldn't be the first time :trollface:

thibault-martinez: You are 100% right and I've had this thought too. We could have `Transaction` and `Statement` to match the database metaphor. Unfortunately you're  also right saying it is controversial. Not entirely sure we can change terms at this point ?

text/0018-bee-bundle/0018-bee-bundle.md
SuperFluffy: If we go for something exotic, then it should be `trits`. `trytes` being defined in terms of `trits` is an extra layer on top of `trits`, so a user would need to keep both `trits` and `trytes` in mind.

I very much like bytes as well, but since then we also need to talk about compact encodings we should probably not do that.

Thoralf-M: Shouldn't the tag also be available to the user?

thibault-martinez: Yes you are right. I'll rephrase, I was referring only to sig/msg :)

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: I tried. Honestly with nested loops it's pretty hard to follow, pipes make it really easy. I'd rather keep them.

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: You're right

text/0018-bee-bundle/0018-bee-bundle.md
thibault-martinez: I renamed. But I omitted parameters on purpose, see comment just after the prototypes, because this is implementation detail.

SuperFluffy:
Incorporated into #20.

While discussing the design of the `Transaction` type, see #3, we realized that a new transaction is not supposed to be create directly. Instead, when constructing a new message, the Bundle is the fundamental object one wants to crate, which in turn splits up its message into several Transactions. A Transaction thus becomes a very simple type. More complicated logic should thus be handled by the Bundle.

This makes it necessary to expand the `Bundle` RFC beyond what is proposed here.

PR #19, opened 2019-10-03
RFC: `bee-pow` crate

Alex6323:
[Rendered](https://github.com/Alex6323/bee-rfcs/blob/rfc-pow/text/0000-bee-pow/0000-bee-pow.md)

### DIFF
--- a/text/0000-bee-pow/0000-bee-pow.md
+++ b/text/0000-bee-pow/0000-bee-pow.md
@@ -1,70 +1,67 @@
-+ Feature name: `bee-pow`
++ Feature name: `bee-pow` 
 + Start date: 2019-10-03
-+ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000) 
 + Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
 
 # Summary
 
-This RFC proposes a dedicated crate to perform proof-of-work (PoW) for a single gossip message, that attempts to become accepted and propagated by the network. 
+IOTA is a permissionless and feeless network which unavoidably makes it vulnerable to spam attacks. To make such
+attacks more costly this RFC proposes a spam protection mechanism based on Proof-of-Work (PoW) where only those
+messages are accepted and propagated by the honest nodes that include a valid nonce value. That nonce serves as a
+proof to show that a predefined amount of computational work has been done upfront either by the publisher itself or
+by some PoW service provider on its behalf.
 
 # Motivation
 
-In order to protect the network from spam each device that wishes to get a message (e.g. a value transaction) propagated by all the nodes needs a way to proove to the network, that it has invested the necessary amount of computational work either by doing the computation by itself, or by externalizing it using some PoW service provider. 
-
-### Background info on PoW
-Cryptographic hash functions are one-way-functions. That means in simple terms that given the output you cannot calculate the input from it. Proof-of-Work works by setting some constraint on the output like a certain number of zeros at the end or the beginning of the hash. It's completely arbitrary. It could be ten `1`s or the sequence `1234567890` because any particular sequence is equally likely. What matters is that your only chance to find an input that satisfies the constraint is guessing/brute-forcing it. You can also easily set the difficulty by making the constraint harder to satisfy (extend the sequence). So to check if PoW was correctly done all a validating node has to do is hash the given message which includes the nonce, and see if the constraint is satisfied. On the other hand the message publisher has to repeat the following cycle many times until it has found a valid nonce:
-* Pick a nonce value (e.g. by simply increasing it, or choose it randomly)
-* Hash the message together with selected nonce
-* Compare if the hash satisfies the constraint given by the majority of the network
-
-This process is called `Mining` and in this analogy the nonce is the nugget.
-
-TODO
-
-Why are we doing this? What use cases does it support? What is the expected
-outcome?
-
-1. Write a summary of the motivation.
-2. List all the specific use cases that your proposal is trying to address. 
-   the software, for example using the "Job story" format:
-
-TODO
+Since IOTA is a permissionless and feeless network it needs methods of protecting itself against spam from
+intentionally or even unintentionally misbehaving network participants (nodes or clients) that otherwise could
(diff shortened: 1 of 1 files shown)


text/0000-bee-pow/0000-bee-pow.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Background info on PoW
+Cryptographic hash functions are one-way-functions. That means in simple terms that given the output you cannot calculate the input from it. Proof-of-Work works by setting some constraint on the output like a certain number of zeros at the end or the beginning of the hash. It's completely arbitrary. It could be ten `1`s or the sequence `1234567890` because any particular sequence is equally likely. What matters is that your only chance to find an input that satisfies the constraint is guessing/brute-forcing it. You can also easily set the difficulty by making the constraint harder to satisfy (extend the sequence). So to check if PoW was correctly done all a validating node has to do is hash the given message which includes the nonce, and see if the constraint is satisfied. On the other hand the message publisher has to repeat the following cycle many times until it has found a valid nonce:
```
tsvisabo: This section could be even shorter and instead include a wiki link to PoW,
we did not come up with PoW

tsvisabo (changes requested):

text/0000-bee-pow/0000-bee-pow.md
Alex6323: you're absolutely right. I will add [this link](https://docs.iota.org/docs/dev-essentials/0.1/concepts/minimum-weight-magnitude).

I hoped that by giving it the title "Background info on PoW", that it would be clear, that we didn't come up with it. 

Do you suggest to remove that section altogether and just put the link there, or should it remain in the RFC, just shortened? @tsvisabo

text/0000-bee-pow/0000-bee-pow.md
```diff
@@ -0,0 +1,332 @@
...
+    H: Hasher<Data = D, Hash = R>,
+    S: Sampler<Data = D>,
+    T: Tester<Hash = R>,
+{
+    hasher: H,
+    sampler: S,
+    tester: T,
+}
+
+impl<H, S, T, D, R> PearlDiver<H, S, T, D, R>
```
wusyong: As long as I pretty love these trait layers, I feel like generics in `PearlDiver` is a little bit out of hand. Would it be better to use trait object in this case? Don't get me wrong, I actually prefer static dispatch over dynamic. Just wanna bring some suggestions and see if it fits in. I would like to hear others' opinion between trait bound and trait objects.

```suggestion
struct PearlDiver<D, R> {
    hasher: Box<dyn Hasher<Data = D, Hash = R>>,
    sampler: Box<dyn Sampler<Data = D>>,
    tester: Box<dyn Tester<Hash = R>>,
}
```

text/0000-bee-pow/0000-bee-pow.md
```diff
@@ -0,0 +1,332 @@
...
+```
+
+Notive that by making `Data` and `Hash` associated types this RFC doesn't make any commitment to the underlying representation which allows for a very flexible design.
+
+## Sampler
+
+The `Sampler` trait allows implementing different types that mutate the input data handed to the hasher on each try. The standard sampler used in IOTA simply starts with a zero nonce and increments it until it finds a valid one. There are, however, other samplers possible, like a random sampler.
+
+```Rust
+pub trait Sampler {
```
semenov-vladyslav: Not sure where to put this comment, so it'll be here. The proposed design is not very well suited for data-parallel (using SIMD instructions) instances. At the moment fastest hashcash and pearldiver implementations use `PCurl`, which is "parallel" Curl capable of hashing `n` data/transactions (where `n=64..256`) at once. Of course, `Sampler::Data` can encapsulate such data-parallelism, but in my opinion it's not very convenient and it's worth baking it into the traits.

text/0000-bee-pow/0000-bee-pow.md
```diff
@@ -0,0 +1,332 @@
...
+    type Data;
+
+    fn next(&mut self, data: &mut Self::Data);
```
semenov-vladyslav: `next` should have `bool` as return type and return `false` in case the search data is exhausted. In real world cases we'll hardly exhaust it but it might be useful in tests and as additional guarantees that PoW doesn't loop infinitely.

text/0000-bee-pow/0000-bee-pow.md
```diff
@@ -0,0 +1,332 @@
...
+Notive that by making `Data` and `Hash` associated types this RFC doesn't make any commitment to the underlying representation which allows for a very flexible design.
+
+## Sampler
+
+The `Sampler` trait allows implementing different types that mutate the input data handed to the hasher on each try. The standard sampler used in IOTA simply starts with a zero nonce and increments it until it finds a valid one. There are, however, other samplers possible, like a random sampler.
+
+```Rust
+pub trait Sampler {
+    type Data;
+
```
semenov-vladyslav: Again, from my experience in hashcash and pearldiver the search data should be prepared in some non-trivial way (mostly, in order to guarantee that parallel ranges do not overlap). I would consider something like this:
```rust
    type Data;
    type SearchData;
    fn prepare(&mut self, data: &Self::Data, search_data: &mut Self::SearchData);
    fn next(&mut self, search_data: &mut Self::SearchData) -> bool;
```

text/0000-bee-pow/0000-bee-pow.md
```diff
@@ -0,0 +1,332 @@
...
+    type Data;
+    type Nonce;
+
+    fn get_nonce(&mut self, data: Self::Data) -> PoWResult<Self::Nonce>);
+}
+```
+whereby
+
+```Rust
+pub type PoWError = Box<dyn std::error::Error>;
```
zesterer: Is there any particular reason that `PoWError` must be dynamically-allocated?

text/0000-bee-pow/0000-bee-pow.md
Alex6323: I agree with you.

text/0000-bee-pow/0000-bee-pow.md
Alex6323: I think not. Any suggestion?

text/0000-bee-pow/0000-bee-pow.md
Alex6323: very good input, thanks. I will think about this more.

Am I correct in assuming that by "prepared in some non-trivial way" you mean conversion from t1b1 to ptrits (vectorizing) ?

PR #20, opened 2019-10-14
RFC: IRI v1.8.1 transactions and bundles

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/rfc-transaction-bundle/text/0020-iri-v1-8-1-transaction-and-bundle/0020-iri-v1-8-1-transaction-and-bundle.md)

Tracking issue: https://github.com/iotaledger/bee/issues/63

This RFC proposes the implementation of the `Transaction` and `Bundle` types to track definition of transactions and bundles as of `iri v1.8.1` (the IOTA reference implementation).

The initial plan was to split [Transaction RFC](https://github.com/iotaledger/bee-rfcs/pull/3) and [Bundle RFC](https://github.com/iotaledger/bee-rfcs/pull/18), we decided to merge them after careful considerations:

+ they are way too tied to be considered apart;
+ we've been moving a lot of logic from Transaction to Bundle;

### DIFF
--- a/text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
+++ b/text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
@@ -500,7 +500,7 @@ struct SignatureInputs {
 }
 ```
 
-### `OutgoingBundleBuilder` and `SealedOutgoingBundleBuilder` struct
+### `OutgoingBundleBuilder` and `SealedOutgoingBundleBuilder`
 
 The `OutgoingBundleBuilder` and `SealedOutgoingBundleBuilder` are more involved compared to their incoming bundle
 sibling. `OutgoingBundleBuilder` allows pushing not yet finished transactions into it. Once all `TransactionBuilder`s
--- a/text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
+++ b/text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
@@ -177,14 +177,13 @@ pub enum TransactionError {
 
 ## Bundle
 
-Transactions are final and bundles, essentially being arrays of transactions, are also final. Once a bundle is created
-and validated, it shouldn't be tempered. For this reason we have a `Bundle` type and a `BundleBuilder` type.
-An instantiated `Bundle` object represents a syntactically and semantically valid IOTA bundle and a `BundleBuilder` is
-the only gateway to a `Bundle` object.
+Bundles are immutable groups of immutable transactions, once a bundle is created and validated, it shouldn't be
+tempered. For this reason we have a `Bundle` type and a `BundleBuilder` type. An instantiated `Bundle` object represents
+a syntactically and semantically valid IOTA bundle and a `BundleBuilder` is the only gateway to a `Bundle` object.
 
 ### Bundle struct
 
-As bundles are final, they shouldn't be modifiable outside of the scope of the bundle module.
+As bundles are immutable, they shouldn't be modifiable outside of the scope of the bundle module.
 
 There is a natural order to transactions in a bundle that can be represented in two ways:
 + each transaction has a `current_index` and a `last_index` and `current_index` goes from `0` to `last_index`, a bundle
@@ -195,15 +194,15 @@ can then simply be represented by a data structure that contiguously keeps the o
 For this reason, we hide this as an implementation detail and instead provide a newtype:
 
 ```rust
-struct Transactions(Vec<Transaction>)
-//struct Transactions(HashMap<Transaction>)
-//struct Transactions(BTreeMap<Transaction>)
+pub struct Transactions(Vec<Transaction>)
+// pub struct Transactions(HashMap<Transaction>)
+// pub struct Transactions(BTreeMap<Transaction>)
 ```
 
 Then the `Bundle` type looks like:
 
 ```rust
-struct Bundle {
+pub struct Bundle {
     transactions: Transactions
 }
 ```
@@ -218,15 +217,16 @@ impl Bundle {
 }
 ```
 
-### BundleBuilder
+### BundleBuilder struct
 
 The `BundleBuilder` offers a simple and convenient way to build bundles:
 
 ```rust
 pub struct BundleBuilder {
-    transactions: Vec<TransactionBuilder>
+    transaction_drafts: TransactionDrafts
 }
 ```
+*`TransactionDrafts` is a constructor for `Transaction`s, but since we don't expose the `transaction_drafts` field publicly, it remains an implementation detail for the end user.*
 
 ```rust
 impl BundleBuilder {    
@@ -530,7 +530,7 @@ fn validate(bundle) -> Result<(), BundleValidationError> {
 
 + A `Bundle` is a fundamental component of the IOTA protocol and must be implemented;
 + There is no more intuitive and simple way to implement a `Bundle` than the one proposed;
-+ Since bundles are final, `BundleBuilder` is mandatory;
++ Since bundles are immutable, `BundleBuilder` is mandatory;
 
 # Unresolved questions
 
@@ -539,4 +539,4 @@ fn validate(bundle) -> Result<(), BundleValidationError> {
 + What should go into a `TransactionError`?
 + Should we use some error libraries?
 + Should this RFC expands a bit more on the M-Bug ? Or give a link ?
-+ Should `Bundle` provide a `.transactions` or a `.at` method ?
++ Should `Bundle` provide a `.next()` method ?
--- a/text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
+++ b/text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
@@ -12,11 +12,11 @@ plain data, are propagated through the IOTA network in transactions. Message tha
 of several transactions and sent one by one.
 
 This RFC proposes a `Transaction` type and a `Bundle` type to represent the transaction and bundle formats used by the
-IOTA Reference Implementation as of release [`iri v1.8.1`]. We also propose the `TransactionBuilder`,
-`IncomingBundleBuilder`, and `OutgoingBundleBuilder` types following the builder pattern to construct both message
-types, respectively. We distinguish between incoming and outgoing bundles because one is concerned with verifying an
-existing message being received, while the other one is concerned with constructing a new message intended to be sent,
-which includes setting a number of fields like the bundle hash.
+IOTA Reference Implementation as of release [`iri v1.8.1`]. To construct these types, this proposal also includes the
+builder patterns `TransactionBuilder`, `IncomingBundleBuilder`, `OutgoingBundleBuilder`, and
+`SealedOutgoingBundleBuilder`. `IncomingBundleBuilder` is concerned with constructing and verifying complete messages
+coming in externally, while `OutgoingBundleBuilder` and its sealed version are for constructing a bundle from scratch to
+be sent to the IOTA network. This includes signing its transactions, setting the bundle hash and other relevant fields.
 
 Useful links:
 
@@ -34,8 +34,8 @@ IOTA is a transaction settlement and data transfer layer for the Internet of Thi
 `Bundle`s of one or more `Transaction`s, which in turn are sent one at a time and are stored in a distributed ledger
 called the *Tangle*. Each transaction encodes data such as sender and receiver addresses, referenced transactions in
 the Tangle, bundle hash, timestamps, and other information required to verify and process each transaction. With
-`Transaction` and `Bundle` representing the units of communication within the IOTA network, we need to be able to
-construct and represent them in memory.
+`Transaction` and `Bundle` representing the units of communication within the IOTA network, one needs to be able
+to construct them and represent in memory.
 
 The current design proposal intends to be as simple as possible, provide an idiomatic Rust interface, and use existing
 IOTA terminology where appplicable. The present types are not intended to be used by end users, but rather as building
@@ -86,9 +86,9 @@ format might change in the future, but for the time being it is not yet understo
 different versions transaction formats should be implemented, or if the network will support several transaction types
 simultaneously.
 
-We thus do not consider generalizations over or interfaces for transactions, but only propose a basic `Transaction`
-type. All mentions of *transactions* in general or the `Transaction` type in particular will be implicitly in reference
-to that format used by `iri v1.8.1`.
+This proposal thus does not consider generalizations over or interfaces for transactions, but only propose a basic
+`Transaction` type. All mentions of *transactions* in general or of the `Transaction` type in particular will be
+implicitly in reference to the format used by `iri v1.8.1`.
 
 [`iri v1.8.1`]: https://github.com/iotaledger/iri/releases/tag/v1.8.1-RELEASE
 
@@ -105,17 +105,16 @@ This RFC proposes the implementation of the following types:
   `TransactionBuilder`s, i.e. transactions that are not yet fully constructed and will be finalized during the
   construction process of the entire `Bundle`.
 
-In this proposal we keep the representation of the various types relatively simple and flat. We assume that the user is
-responsible for pushing the appropriate number of transaction drafts into a bundle builder and don't encode this on the
-type level. During the construction phase we check that all transactions are consistent and adhere to IOTA convention,
-but we don't, for example:
+This proposal keeps the representation of the various types flat. It is assumed that the user is responsible for pushing
+the appropriate number of transaction drafts into a bundle builder. During the construction phase, all transactions are
+checked for consistency and for adherence to IOTA conventions. This proposal does not however:
 
-+ attempt to represent different the different types of transactions on the type level;
++ attempt to represent different the different types of transactions at the type level;
 + allow the bundle builders to push extra transactions into their collection automatically to encode a signature of
   a certain size.
 
-The present proposal is meant as a building block for client facing code to provide higher level types that provide
-more convenient and implicit ways of constructing transactions and bundles.
+The proposed types are meant as building blocks for higher level abstractions in client facing code to provide more
+convenient and implicit ways of constructing transactions and bundles.
 
 A transaction is 8019 trits large in total, out of which 6551 trits are available to store a payload. The payload is
 defined to either hold a signature fragment or a message fragment. Since it has a limited size, a user often needs more
@@ -218,12 +217,12 @@ impl Transaction {
 }
 ```
 
-We treat the types of the `Transaction` struct's fields as opaque newtypes for now so that we can flesh them out during
-the implementation phase or in future RFCs. Because this RFC implements the transaction format as of iri v1.8.1, the
-newtypes shall be constructed from byte slices of a certain length matching that of the reference implementation.
-Conversion methods shall only verify that the passed byte slices (or fixed-size arrays or vectors of bytes) are of
-appropriate length and that each contained byte correctly encodes a balanced trit (this is also refered to as `T1B1`
-binary-coded ternary encoding).
+The fields in the `Transaction` struct are opaque newtypes so that they can be fleshed out during the implementation
+phase or in future RFCs without requireing breaking changes. Because this RFC implements the transaction format as of
+iri v1.8.1, the newtypes shall be constructed from byte slices of a certain length matching that of the reference
+implementation. Conversion methods shall only verify that the passed byte slices (or fixed-size arrays or vectors of
+bytes) are of appropriate length and that each contained byte correctly encodes a balanced trit (this is also refered to
+as `T1B1` binary-coded ternary encoding).
 
 ```rust
 pub struct SignatureOrMessageFragment([u8; 6561]);
@@ -237,9 +236,9 @@ pub struct TransactionHash([u8; 243]);
 pub struct Nonce([u8; 81]);
 ```
 
-Below is an example implementation for the `Tag` type. We only consider conversion from byte slices, `&[u8]`. Checked
-conversions ensuring that each byte encodes `-1`, `0`, or `+1` are implemented in terms of the `TryFrom` trait,
-while we also provide `Tag::from_unchecked` as an unchecked faster constructor method.
+Below is an example implementation for the `Tag` type. Only conversions from byte slices, `&[u8]`, are considered.
+Checked conversions ensuring that each byte encodes `-1`, `0`, or `+1` are implemented in terms of the `TryFrom` trait,
+while a `Tag::from_unchecked` function is provided as an unchecked faster constructor method.
 
 ```rust
 enum TagError {
@@ -327,10 +326,11 @@ pub struct TransactionBuilder {
 
 `TransactionBuilder`'s setter methods are implemented using generic type parameters `T: TryInto<{FieldType}>` to provide
 some convenience when setting fieds from byte slices. The `TryFrom` implementations for each field type ensure that the
-byte slices only encode correct trits. We allow to the checks when setting the fields from trusted sources by explicitly
-constructing the target object. For example, when setting the `tag` field on a `builder: TransactionBuilder` object with
-`trusted_bytes: &[u8]`, calling `builder.tag(Tag::from_unchecked(trusted_bytes))` sets the `tag` field to a value that
-was not verified to only contained valid bytes.
+byte slices only encode correct trits. If the data comes from a trusted source, the checks can be circumvented by
+explicitly constructing the target object and that one in the setter. For example, when setting the `tag` field on
+a `builder: TransactionBuilder` object with `trusted_bytes: &[u8]`, calling
+`builder.tag(Tag::from_unchecked(trusted_bytes))` sets the `tag` field to a value that was not verified to only
+contained valid bytes.
 
 ```rust
 enum TransactionBuilderError {
@@ -386,8 +386,8 @@ impl TransactionBuilder {
 ### `Error` types
 
 The `TransactionBuilderError` enum contains variants for each of its fields, encoding that an error can occur when
-setting any of them. In addition, it shall contain errors that can occur during verification or construction. We leave
-the specification of these extra error variants to the implementation phase.
+setting any of them. In addition, it shall contain errors that can occur during verification or construction. The
+specification of the concrete error types are left to be elaborated in the implementation phase.
 
 ## `Bundle`
 
@@ -408,11 +408,11 @@ There is a natural order to transactions in a bundle that can be represented in 
 
 + each transaction has a `current_index` and a `last_index`. `current_index` goes from `0` to `last_index`. A bundle can
   then simply be represented by a data structure that contiguously keeps the order like `Vec`;
-+ each transaction is chained to the next one through its `trunk` which means we can consider data structures like
-  `HashMap` or `BTreeMap`;
++ each transaction is chained to the next one through its `trunk` hash, which means one can consider using
+  a datastructure like `HashMap`.
 
-For the purpose of this RFC, we opt for the simplest implementation in terms of `Vec` but hide the details behind
-a newtype and leave the details of the underlying datastructure open to be changed in the future.
+This RFC opts for using the simplest implementation in terms of `Vec`, but hides the implementation details behind
+a newtype. This way, the underlying datastructures can be changed in the future without breaking dependent code.
 
 ```rust
 pub struct Transactions(Vec<Transaction>)
@@ -531,7 +531,7 @@ pub struct SealedOutgoingBundleBuilder {
 struct TransactionBuilders(Vec<TransactionBuilder>);
 ```
 
-As with `Transactions`, we make `TransactionBuilders` an opaque newtype that for the time being wraps a vector of
+As with `Transactions`, `TransactionBuilders` are an opaque newtype that for the time being wraps a vector of
 `TransactionBuilder`s.
 
 ```rust
@@ -882,9 +882,9 @@ impl SealedOutgoingBundleBuilder {
 
 ## How this is used
 
-In this section, we describe the algorithms needed to build a `Bundle`. The work flow depends on whether one is
-receiving a bundle (and its constituent transactions), or creating a new one to send it. `IncomingBundleBuilder` encodes
-receipt of a bundle, while `OutgoingBundleBuilder` encodes sending.
+This section shows how to construct a `Bundle`. The work flow depends on whether one is receiving a bundle (and its
+constituent transactions), or creating a new one to send it. `IncomingBundleBuilder` encodes receipt of a bundle, while
+`OutgoingBundleBuilder` encodes sending.
 
 ### Workflow of `IncomingBundleBuilder`
 
@@ -949,12 +949,12 @@ let bundle = outgoing_bundle_builder
 
 + The design of the builders currently take ownership and return `Self`. This is nice for chaining calls, but might
   suboptimal for performance. Is it more useful to take borrows, either `&Self` or `&mut Self`, instead?
-+ `getTransactionsToApprove` function: the function executing proof of work on `SealedOutgoingBundleBuilder` is talking about
-  getting tip and branch from the mentionede function. Should it be part of this RFC or assumed external?
++ `getTransactionsToApprove` function: the function executing proof of work on `SealedOutgoingBundleBuilder` is talking
+  about getting tip and branch from the mentionede function. Should it be part of this RFC or assumed external?
 + Should the `chain` function on `SealedOutgoingBundleBuilder` be renamed to `attach`? Is there a better name?
-+ When validating and building the `SealedOutgoingBundleBuilder`, does it make sense to validate withdrawal transactions one
-  more time? Should there be an extra type, e.g. `SealedSignedOutgoingBundleBuilder`, where we encode on the type level
-  that the transaction is verified?
++ When validating and building the `SealedOutgoingBundleBuilder`, does it make sense to validate withdrawal transactions
+  one more time? Should there be an extra type, e.g. `SealedSignedOutgoingBundleBuilder`, where we encode on the type
+  level that the transaction is verified?
 
 # Blockers
 
--- a/text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
+++ b/text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
@@ -14,11 +14,9 @@ of several transactions and sent one by one.
 This RFC proposes a `Transaction` type and a `Bundle` type to represent the transaction and bundle formats used by the
 IOTA Reference Implementation as of release [`iri v1.8.1`]. We also propose the `TransactionBuilder`,
 `IncomingBundleBuilder`, and `OutgoingBundleBuilder` types following the builder pattern to construct both message
-types, respectively. We distinguish between incoming and outgoing bundles because one is concerned with verifying the
-veracity of an existing message being received, while the other one is concerned with constructing a new message
-intended to be sent, which includes setting a number of fields like the bundle hash.
-
-By analogy with IP fragmentation, a bundle corresponds to a packet, while transactions correspond to fragments.
+types, respectively. We distinguish between incoming and outgoing bundles because one is concerned with verifying an
+existing message being received, while the other one is concerned with constructing a new message intended to be sent,
+which includes setting a number of fields like the bundle hash.
 
 Useful links:
 
@@ -32,12 +30,56 @@ Useful links:
 
 # Motivation
 
-IOTA is a transaction settlement and data transfer layer for IoT (the Internet of Things). Messages on the network are
-`Bundle`s of individual `Transaction`s, which in turn are sent one at a time and are stored in a distributed ledger
-called the *Tangle*. Each `Transaction` encodes data such as sender and receiver addresses, referenced transactions in
-the Tangle, `Bundle` hash, timestamps, and other information required to verify and process each transaction. With
+IOTA is a transaction settlement and data transfer layer for the Internet of Things. Messages on the network are
+`Bundle`s of one or more `Transaction`s, which in turn are sent one at a time and are stored in a distributed ledger
+called the *Tangle*. Each transaction encodes data such as sender and receiver addresses, referenced transactions in
+the Tangle, bundle hash, timestamps, and other information required to verify and process each transaction. With
 `Transaction` and `Bundle` representing the units of communication within the IOTA network, we need to be able to
 construct and represent them in memory.
+
+The current design proposal intends to be as simple as possible, provide an idiomatic Rust interface, and use existing
+IOTA terminology where appplicable. The present types are not intended to be used by end users, but rather as building
+blocks for higher level APIs. As such, the different kinds of transactions are not encoded as Rust types, and the bundle
+builders do not contain any logic to implicitly push more transactions into their stack. For a bundle to be buildable,
+all required transactions have to be present when validating and building. Otherwise the build will fail.
+
+As an example, a simple `Bundle` that intends to transfer funds, withdrawing tokens from one address and depositing them
+in another would be constructed like this:
+
+```rust
+let mut outgoing_bundle_builder = OutgoingBundleBuilder::new();
+
+let mut withdrawing_tx_builder = TransactionBuilder::new();
+let mut depositing_tx_builder = TransactionBuilder::new();
+
+// The transactions here only need to set addresses, values, and tags. All other fields, such as the signature
+// of the withdrawing transaction, will be filled in by the bundle builder.
+withdrawing_tx_builder
+    .address(address_providing_funds)
+    .value(value_to_transfer)
+    .tag(custom_tag);
+
+depositing_tx_builder
+    .address(address_receiving_funds)
+    .value(value_to_transfer)
+    .tag(custom_tag);
+
+let outgoing_bundle = outgoing_bundle_builder
+    .push(withdrawing_tx_builder)
+    .push(depositing_tx_builder)
+    // Seal the bundle, adding indices to its transactions and assigning a bundle hash. This also prevents
+    // pushing more transactions into the bundle.
+    .seal()?
+    // Withdrawal transactions are signed, providing some signature scheme, seed, and wallet. If none are present, this
+    // step can be skipped.
+    .sign(signature_scheme, seed, wallet)?
+    // Attach to the tangle, chain transactions to each other by performing proof of work and calculating transaction hashes via the sponge.
+    .chain_and_attach(trunk_tip, branch_tip, proof_of_work, sponge)?
+    // Check if all transactions contained in the sealed builder are consistent.
+    .validate(sponge)?
+    // Construct the final `Bundle`.
+    .build()?;
+```
 
 At the time of this RFC, the transaction format used by the IOTA network is defined by [`iri v1.8.1`]. The transaction
 format might change in the future, but for the time being it is not yet understood how a common interface between
@@ -47,25 +89,6 @@ simultaneously.
 We thus do not consider generalizations over or interfaces for transactions, but only propose a basic `Transaction`
 type. All mentions of *transactions* in general or the `Transaction` type in particular will be implicitly in reference
 to that format used by `iri v1.8.1`.
-
-A transaction is 8019 trits large in total, out of which 6551 trits are available to store a payload. The payload is
-defined to either hold a signature fragment or a message fragment. Since it has a limited size, a user often needs more
-than one transaction to execute an operation. For example, signatures with [security level 2 or 3](Security levels)
-don't fit in a single transaction, and user-provided messages may exceed the maximum payload capacity so they need to be
-fragmented across multiple transactions. Moreover, because the total amount of tokens stored in the ledger has to stay
-constant, *input transactions* (which are called thus because an address is *put into* a transfer as a source of tokens,
-thus removing them from the address) have to always be matched with *output transactions* such that their total value is
-zero. For these reasons, transactions have to be processed as a whole in groups called bundles. A bundle is an atomic
-operation in the sense that either all or none of its transactions are accepted by the network. Even single transactions
-are propagated through the network within a bundle.
-
-The `Transaction` and `Bundle` types are intended to be final and immutable. This contract is enforced by not exposing
-any methods that allow manipulating their fields directly. `Transaction`s are intended to be constructed through
-the `TransactionBuilder` builder pattern, while `Bundle`s can be created either via the `IncomingBundleBuilder` or
-`OutgoingBundleBuilder`. The `IncomingBundleBuilder` takes in existing `Transaction`s, compiles them, and verifies
-their veracity before constructing a `Bundle` type. The `OutgoingBundleBuilder` takes in `TransactionBuilder`s,
-and is primarily responsible for filling in all information that can only be set in the actual finalization
-of the outgoing message into a `Bundle`, such as the bundle hash.
 
 [`iri v1.8.1`]: https://github.com/iotaledger/iri/releases/tag/v1.8.1-RELEASE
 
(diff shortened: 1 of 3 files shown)


text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,572 @@
...
+[iotaledger/bee-rfcs#18](https://github.com/iotaledger/bee-rfcs/pull/18),
+[iotaledger/bee-rfcs#20](https://github.com/iotaledger/bee-rfcs/pull/20)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+## Transaction
+
+In the context of IOTA, a `Bundle` is a message that is shared over the network, and a `Transaction` is the fundamental
+unit that a `Bundle` is constructed from. To use TCP as a well known analogy, a `Bundle` corresponds to a message, and a `Transaction` corresponds to a packet.
```
SuperFluffy: We should ~~call~~ describe `Bundle` a `Stream` instead, to get the TCP analogy right.

samuel-rufi: I wouldn't introduce new terms in this RFC for bundles/transactions, we should stick to them as we are used to. Introducing new, more exact terms could be done within Coordicide. I'm actually in for rethinking  some of our terms, as well as rethinking the structure of transaction and bundle. I like the TCP analogy, Transaction should be the fundamental packet layer, and we should keep in mind, it should be **independent** of the overlying bundle/message layer.

text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,569 @@
...
++ RFC PR: [iotaledger/bee-rfcs#3](https://github.com/iotaledger/bee-rfcs/pull/3),
+[iotaledger/bee-rfcs#18](https://github.com/iotaledger/bee-rfcs/pull/18),
+[iotaledger/bee-rfcs#20](https://github.com/iotaledger/bee-rfcs/pull/20)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+## Transaction
+
+In the context of IOTA, a `Bundle` is a message that is shared over the network, and a `Transaction` is the fundamental
```
samuel-rufi: This sentence seems bit spongy. Instead of writing
> a `Bundle` is a message that is shared over the network

maybe say something like "messages are split across several packets when they don't fit into one packet" and then directly lead over to your TCP analogy.

text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,569 @@
...
+unit that a `Bundle` is constructed from. To use TCP as a well known analogy, a `Bundle` corresponds to a message, and a `Transaction` corresponds to a packet.
+
+This RFC proposes a `Transaction` type to represent the transaction format used by the IOTA Reference Implementation as
+of version [`iri v1.8.1`], commit `e1776fbad5d90df86a26402f9025e4b0b2ef7d3e`. The `Transaction` type is implemented as a
+struct and is intended to be a public interface. It thus comes with a number of opaque types for its fields that we
```
samuel-rufi: What do you mean with public interface? A reader could be confused by the non mutability of a transaction and the "intended to be a public interface"

text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,569 @@
...
+    pub fn from_slice(stream: &[u8]) -> Result<Self, TransactionDraftError> {
+        unimplemented!()
+    }
+
+    /// Verify that all fieleds are set and build the `Transaction`.
+    pub fn build(self) -> Result<Transaction, TransactionDraftError> {
+        unimplemented!()
+    }
+
+    pub fn address(&mut self, address: Address) -> &mut Self {
```
samuel-rufi: We should use the `set` prefix for setters as described in RFC 344: [https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis)

text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,569 @@
...
+## Bundle
+
+The smallest communication unit in the IOTA protocol is the transaction. Everything, including payment settlements
+and/or plain data, is propagated through the IOTA network in transactions.
+
+A transaction is `8019` trits and the payload (i.e. `sig_or_msg`) is `6561` trits. This payload holds a signature in
+case of a payment settlement and plain data otherwise. Since it has a limited size, a user often needs more than one
+transaction to fulfil their operation, for example signatures with security level `2` or `3` don't fit in a single
+transaction and user-provided data may exceed the allowance so they need to be fragmented across multiple transactions.
+Moreover, a value transaction doesn't make sense on its own because it would change the total amount of the ledger so
```
samuel-rufi: Maybe say also that the term "value transaction" shouldn't be used since its not correct. Instead recommend a term like "payment settlement", since it doesn't come across with our `Transaction` struct.

text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,569 @@
...
+We also propose a `TransactionDraft` type that follows the builder pattern. It is intended to be an implementation
+detail and only be used by a `Bundle` builder type and not exposed publicly.
+
+## Bundle
+
+The smallest communication unit in the IOTA protocol is the transaction. Everything, including payment settlements
+and/or plain data, is propagated through the IOTA network in transactions.
+
+A transaction is `8019` trits and the payload (i.e. `sig_or_msg`) is `6561` trits. This payload holds a signature in
+case of a payment settlement and plain data otherwise. Since it has a limited size, a user often needs more than one
```
samuel-rufi: I think this needs to be defined more exact. In case of a payment settlement, there exist transactions (in the payment settlement bundle), which don't contain any signature. So this isn't fully correct: 
> This payload holds a signature in case of a payment settlement and plain data otherwise.

See following bundle [https://thetangle.org/bundle/XBKKFHIXYSLEDDCYEEQCCOOKSJVWEOQQGBFAOAJCHGORE9VYEKVCK9OINCHKJGFESZNPZYASDECUDYIZD](https://thetangle.org/bundle/XBKKFHIXYSLEDDCYEEQCCOOKSJVWEOQQGBFAOAJCHGORE9VYEKVCK9OINCHKJGFESZNPZYASDECUDYIZD)

and following "value" transaction [https://thetangle.org/transaction/YNFATJGQBNIKNBKJCPNIYUJCUFVKFHMFH9KGTEHWPDSEPM9HNTWUKCLPQSKQZAEHIRIB9CRFTFQJZ9999](https://thetangle.org/transaction/YNFATJGQBNIKNBKJCPNIYUJCUFVKFHMFH9KGTEHWPDSEPM9HNTWUKCLPQSKQZAEHIRIB9CRFTFQJZ9999)

see also the "zero value" tx, which contains signature fragments

[https://thetangle.org/transaction/HCKTCKDLW9ILZNYTSEBPTPDPWKIOHNOMLBMYS9EEBIEUJHREAYTUJNMNAOFVVREBOSPMUYLFLPSSA9999](https://thetangle.org/transaction/HCKTCKDLW9ILZNYTSEBPTPDPWKIOHNOMLBMYS9EEBIEUJHREAYTUJNMNAOFVVREBOSPMUYLFLPSSA9999)

text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,569 @@
...
+transaction to fulfil their operation, for example signatures with security level `2` or `3` don't fit in a single
+transaction and user-provided data may exceed the allowance so they need to be fragmented across multiple transactions.
+Moreover, a value transaction doesn't make sense on its own because it would change the total amount of the ledger so
+it has to be paired with other complementary transactions that together will balance the total value to zero.
+
+For these reasons, transactions have to be processed as a whole, in groups called bundles. A bundle is an atomic
+operation in the sense that either all or none of its transactions are accepted by the network. Even single
+transactions are propagated through the network within a bundle making it the only confirmable communication unit of
```
samuel-rufi: Isn't the term `confirmation/confirmed` bit meaningless in an open network? Also Paul Handy, as well as CfB - if I remember correctly - don't like it.

text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,569 @@
...
++ [Structure of a transaction](https://docs.iota.org/docs/dev-essentials/0.1/references/structure-of-a-transaction)
++ [Structure of a bundle](https://docs.iota.org/docs/dev-essentials/0.1/references/structure-of-a-bundle)
+
+# Motivation
+
+## Transaction
+
+IOTA is a transaction settlement and data transfer layer for IoT (the Internet of Things). Messages on the network are
+`Bundle`s of individual `Transaction`s, which in turn are sent one at a time and are stored in a distributed ledger
+called the *Tangle*. Each `Transaction` encodes data such as sender and receiver addresses, reference nodes in the
```
samuel-rufi: Replace "reference nodes" with something like "referenced transactions".

text/0000-transaction-bundle/0000-transaction-bundle.md
```diff
@@ -0,0 +1,569 @@
...
+    tag: Option<Tag>,
+    attachment_timestamp: Option<AttachmentTimestamp>,
+    attachment_timestamp_lower_bound: Option<AttachmentTimestampLowerbound>,
+    attachment_timestamp_upper_bound: Option<AttachmentTimestampUpperbound>,
+    nonce: Option<Nonce>
+}
+
+impl TransactionDraft {
+    /// Create a `TransactionDraft` from a reader.
+    pub fn from_reader<R: std::io::Read>(reader: R) -> Result<Self, TransactionDraftError> {
```
samuel-rufi: Not sure if we need the `from_reader()` and the `from_slice()` here.

text/0000-transaction-bundle/0000-transaction-bundle.md
SuperFluffy: By convention this is not done in builders, as is noted in the text you have linked:

> Note that this convention is about getters/setters on ordinary data types, not on builder objects. The naming conventions for builder methods are still open.

text/0000-transaction-bundle/0000-transaction-bundle.md
thibault-martinez: Why is it not correct ? "value transaction" is extremely used in the IOTA ecosystem to talk about a transactions that have a value != 0. I find the term "payment settlement" way less approchable.

text/0000-transaction-bundle/0000-transaction-bundle.md
samuel-rufi: True, then maybe we review this at a later point in time once the convention is clear.

text/0000-transaction-bundle/0000-transaction-bundle.md
thibault-martinez: I'll rephrase.

text/0000-transaction-bundle/0000-transaction-bundle.md
samuel-rufi: Just because its used, doesn't mean we should go with it. It seems confusing and not correct. changing earlier is easier than later

text/0000-transaction-bundle/0000-transaction-bundle.md
thibault-martinez: I don't think it's meaningless in the sense it's being used. The only unit to which consensus can apply is the bundle.

text/0000-transaction-bundle/0000-transaction-bundle.md
thibault-martinez: But it's absolutely correct semantically. It's a transaction, with a value. We can't use "payment settlement" for a single transaction.

samuel-rufi: As you said,  a value transaction doesn't make sense on its own. So why do you still want to call it value transaction? A value **transfer/settlement/transaction** needs the help of multiple transactions. Seems like the term value transaction for a single transaction doesn't fit.

text/0000-transaction-bundle/0000-transaction-bundle.md
thibault-martinez: It's been removed.

oopsmonk:
> [Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/rfc-transaction-bundle/text/0000-transaction-bundle/0000-transaction-bundle.md)

wrong link.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,489 @@
...
+## Transaction
+
+IOTA is a transaction settlement and data transfer layer for IoT (the Internet of Things). Messages on the network are
+`Bundle`s of individual `Transaction`s, which in turn are sent one at a time and are stored in a distributed ledger
+called the *Tangle*. Each `Transaction` encodes data such as sender and receiver addresses, referenced transactions in
+the Tangle, `Bundle` hash, timestamps, and other information required to verify and process each transaction. With the
+transaction being the fundamental unit that is sent over the network, we need to represent it in memory.
+
+At the time of this RFC, the transaction format used by the IOTA network is defined by [release v1.8.1 of the IOTA
+Reference Implementation, IRI](`iri v1.8.1`), commit `e1776fbad5d90df86a26402f9025e4b0b2ef7d3e`. The transaction format
```
oopsmonk:
```suggestion
Reference Implementation, IRI](`iri v1.8.1`), commit [e1776fbad5d90df86a26402f9025e4b0b2ef7d3e](https://github.com/iotaledger/iri/tree/e1776fbad5d90df86a26402f9025e4b0b2ef7d3e). The transaction format
```

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,489 @@
...
+name, description, and size (in trits) is summarized in the table below, in their order of appearance in a transaction.
+
+| Name                              | Description                                            | Size (in trits) |
+| ---                               | ---                                                    | ---             |
+| `signature_or_message_fragment`   | contains the signature of the transfer                 |                 |
+|                                   | or user-defined message data                           | 6561            |
+| `address`                         | receiver (output) if value > 0,
+|                                   | or sender (input) if value < 0                         | 243             |
+| `value`                           | the transferred amount in IOTA                         | 81              |
+| `obsolete_tag`                    | another arbitrary user-defined tag                     | 81              |
```
oopsmonk: don't think `obsolete_tag` is defined by the user, it's related to the M bug and will be changed in finalization.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,489 @@
...
+| `timestamp`                       | the time when the transaction was issued               | 27              |
+| `current_index`                   | the position of the transaction in its bundle          | 27              |
+| `last_index`                      | the index of the last transaction in the bundle        | 27              |
+| `bundle`                          | the hash of the bundle essence                         | 243             |
+| `trunk`                           | the hash of the first transaction referenced/approved  | 243             |
+| `branch`                          | the hash of the second transaction referenced/approved | 243             |
+| `tag`                             | arbitrary user-defined value                           | 81              |
```
oopsmonk:
```suggestion
| `tag`                             | arbitrary user-defined tag                           | 81              |
```

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,489 @@
...
+}
+```
+
+```rust
+impl BundleBuilder {    
+    pub fn calculate_hash(&self) {
+      unimplemented!()
+    }
+
+    pub fn finalise(&self) {
```
oopsmonk: for consistency
```suggestion
    pub fn finalize(&self) {
```

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,489 @@
...
+| | | | fragments = sign(seed, input.index, input.security, transaction.bundle)
+| | | | for fragment in fragments
+| | | | | bundle.at(current_index).signature = fragment
+| | | | | current_index = current_index + 1
+| | else
+| | | current_index = current_index + 1
+```
+
+*Since signature size depends on the security level, a single signature can spread out to up to 3 transactions.
+`inputs` is an object that contain all unused addresses of a seed with a sufficient balance.*
```
oopsmonk:
```suggestion
`inputs` is an object that contains all unused addresses of a seed with a sufficient balance.*
```

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,489 @@
...
+
+#### Proof of Work
+
+*Client side operation.*
+
+Proof of Work (PoW) allows your transactions to be accepted by the network. On the IOTA network, PoW is only a rate
+control mechanism. Doing PoW on a bundle means doing PoW on each of its transactions and setting trunks and branch
```
oopsmonk:
```suggestion
PoW is meant to prevent Sybil attack. Doing PoW on a bundle means doing PoW on each of its transactions and setting trunks and branch
```

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
thibault-martinez: Not sure what commit you're reviewing but I simply removed mention of the commit hash. It was not necessary since already included in the IRI version.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
thibault-martinez: Thanks

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
thibault-martinez: Will do

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
thibault-martinez: PoW is a rate control mechanism, not an anti-Sybil protection. Mana is.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,489 @@
...
+their operation, for example signatures with security level `2` or `3` don't fit in a single transaction and
+user-provided message may exceed the allowance so they need to be fragmented across multiple transactions. Moreover,
+input/output transactions doesn't make sense on their own because they would change the total amount of the ledger so
+they have to be paired with other input/output transactions that together will balance the total value to zero.
+
+For these reasons, transactions have to be processed as a whole, in groups called bundles. A bundle is an atomic
+operation in the sense that either all or none of its transactions are accepted by the network. Even single
+transactions are propagated through the network within a bundle.
+
+By analogy with TCP, a bundle corresponds to a stream, and a transaction corresponds to a packet.
```
Alex6323: Are we expecting RFC readers to be knowledgeable in TCP to understand this analogy? Apart from that the analogy is bad because a bundle is not a stream (which insinuates potential endlessness), it's much more similar to a container.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,489 @@
...
+| `trunk`                           | the hash of the first transaction referenced/approved  | 243             |
+| `branch`                          | the hash of the second transaction referenced/approved | 243             |
+| `tag`                             | arbitrary user-defined value                           | 81              |
+| `attachment_timestamp`            | the timestamp for when Proof-of-Work is completed      | 27              |
+| `attachment_timestamp_lowerbound` | *not specified*                                        | 27              |
+| `attachment_timestamp_upperbound` | *not specified*                                        | 27              |
+| `nonce`                           | the Proof-of-Work nonce of the transaction             | 81              |
+
+Each transaction is uniquely identified by its *transaction hash*, which is calculated based on all fields of the
+transaction. Note that the transaction hash is not part of the transaction.
```
Alex6323: Just for the sake of discussion:
Conceptually the `Nonce` shouldn't be part of the transaction either, because it's part of a spam protection mechanism. 
Again, just a point to talk about for future design considerations.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
oopsmonk: It's the same hash but with link on doc. If you want to put the commit hash here, better have a link instead of text only.

thibault-martinez: I totally removed the commit hash.

thibault-martinez: Ahhhh sorry I didn't see the line! There was two place with the exact same hash and I removed the other! Let me do that right now.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
thibault-martinez: I rephrased, thanks

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
thibault-martinez: Janis was using `value` as content, not as field name, I think it's fine.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
thibault-martinez: I get your point. But this RFC is about IRI 1.8.1 transaction. If nonce is not part of the transaction then you somehow have to consider global identities and opens up Sybil attacks. I believe this is atm a research topic.

wusyong (approved):
LGTM. But maybe we could just leave documentation on code blocks to implement phase.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+The fundamental communication unit in the IOTA protocol is the transaction. Messages, including payment settlements and
+plain data, are propagated through the IOTA network in transactions. Message that are too large are split up into bundles
+of several transactions and sent one by one.
+
+This RFC proposes a `Transaction` type and a `Bundle` type to represent the transaction and bundle formats used by the
+IOTA Reference Implementation as of release [`iri v1.8.1`]. We also propose the `TransactionBuilder`,
```
Alex6323: * I think for consistency it's better to always stick to 3rd person ("this RFC proposes ...", "it also proposes", ...) rather than 1st person plural ("we propose ...") 
* Message[s] that are too large ...

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
++ RFC PR: [iotaledger/bee-rfcs#3](https://github.com/iotaledger/bee-rfcs/pull/3),
+  [iotaledger/bee-rfcs#18](https://github.com/iotaledger/bee-rfcs/pull/18),
+  [iotaledger/bee-rfcs#20](https://github.com/iotaledger/bee-rfcs/pull/20)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+The fundamental communication unit in the IOTA protocol is the transaction. Messages, including payment settlements and
+plain data, are propagated through the IOTA network in transactions. Message that are too large are split up into bundles
+of several transactions and sent one by one.
```
Alex6323: I have a proposal for that paragraph which I find more clear: 

Messages in the IOTA network can either represent payment settlements or serve as plain data carriers. In IOTA terminology messages are referred to as 'Bundles', which - depending on the message size - consist of one or more transactions. A transaction is the smallest communication unit. It has a fixed size, while messages in general can have arbitrary size. (maybe support this explanation with an example, e.g. security 3 transfer)

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+
+This RFC proposes a `Transaction` type and a `Bundle` type to represent the transaction and bundle formats used by the
+IOTA Reference Implementation as of release [`iri v1.8.1`]. We also propose the `TransactionBuilder`,
+`IncomingBundleBuilder`, and `OutgoingBundleBuilder` types following the builder pattern to construct both message
+types, respectively. We distinguish between incoming and outgoing bundles because one is concerned with verifying an
+existing message being received, while the other one is concerned with constructing a new message intended to be sent,
+which includes setting a number of fields like the bundle hash.
```
Alex6323: I think fields are set in both cases. The differentiating factor is, that for the outgoing bundle we actually have to create/calculate those fields for the first time while for the incoming bundle those fields can just be read from the wire bytes.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
++ [What is a transaction?](https://docs.iota.org/docs/getting-started/0.1/introduction/what-is-a-transaction)
++ [What is a bundle?](https://docs.iota.org/docs/getting-started/0.1/introduction/what-is-a-bundle)
++ [Bundles and transactions](https://docs.iota.org/docs/dev-essentials/0.1/concepts/bundles-and-transactions)
++ [Structure of a transaction](https://docs.iota.org/docs/dev-essentials/0.1/references/structure-of-a-transaction)
++ [Structure of a bundle](https://docs.iota.org/docs/dev-essentials/0.1/references/structure-of-a-bundle)
++ [Security levels](https://docs.iota.org/docs/dev-essentials/0.1/references/security-levels)
+
+# Motivation
+
+IOTA is a transaction settlement and data transfer layer for the Internet of Things. Messages on the network are
```
Alex6323: we should not restrict ourselves to IoT. IOTA should also be for the classic Internet.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
++ `OutgoingBundleBuilder` and `SealedBundleBuilder`, two builder patterns working in tandem to create a `Bundle` from
+  `TransactionBuilder`s, i.e. transactions that are not yet fully constructed and will be finalized during the
+  construction process of the entire `Bundle`.
+
+In this proposal we keep the representation of the various types relatively simple and flat. We assume that the user is
+responsible for pushing the appropriate number of transaction drafts into a bundle builder and don't encode this on the
+type level. During the construction phase we check that all transactions are consistent and adhere to IOTA convention,
+but we don't, for example:
+
++ attempt to represent different the different types of transactions on the type level;
```
Alex6323: remove one 'different'

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+thus removing them from the address) have to always be matched with *output transactions* such that their total value is
+zero. For these reasons, transactions have to be processed as a whole in groups called bundles. A bundle is an atomic
+operation in the sense that either all or none of its transactions are accepted by the network. Even single transactions
+are propagated through the network within a bundle.
+
+The `Transaction` and `Bundle` types are intended to be final and immutable. This contract is enforced by not exposing
+any methods that allow manipulating their fields directly. `Transaction`s are intended to be constructed through
+the `TransactionBuilder` builder pattern, while `Bundle`s can be created either via the `IncomingBundleBuilder` or
+`OutgoingBundleBuilder`. The `IncomingBundleBuilder` takes in existing `Transaction`s, compiles them, and verifies
+their veracity before constructing a `Bundle` type. The `OutgoingBundleBuilder` takes in `TransactionBuilder`s,
```
Alex6323: i'm not sure 'veracity' is the correct term here. Sounds more like a metaphysical term. What about 'validity'?

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+    branch: TransactionHash,
+    tag: Tag,
+    attachment_timestamp: Timestamp,
+    attachment_timestamp_lower_bound: Timestamp,
+    attachment_timestamp_upper_bound: Timestamp,
+    nonce: Nonce
+}
+```
+
+The `Transaction` type only contains two constructor methods getter methods for read-only borrows of its fields.
```
Alex6323: - two constructor functions and some getter methods for read-only borrows of its fields

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+
+    pub fn value(&self) -> &Value {
+        &self.value
+    }
+}
+```
+
+We treat the types of the `Transaction` struct's fields as opaque newtypes for now so that we can flesh them out during
+the implementation phase or in future RFCs. Because this RFC implements the transaction format as of iri v1.8.1, the
+newtypes shall be constructed from byte slices of a certain length matching that of the reference implementation.
```
Alex6323: a Java byte is an 'i8' in Rust, not an 'u8'. So if we want to match the reference implementation we should use 'i8' in this RFC as the proper equivalent. It also simplifies calculations as we tell the compiler about how we expect the bits in the byte to be interpreted (0xFF as -1, not as 255).

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
bingyanglin: I think the term IOT already includes classic internet and has much boarder meaning.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+        let builder_iterator = self.iter_from_head();
+
+        // Set the head transaction
+        let head_transaction = builder_iterator.next().ok_or("builder does not contain any transaction builders")?;
+        head_transaction.trunk(trunk_tip);
+        head_transaction.branch(branch_tip);
+
+        head_transaction.attachment_timestamp = timestamp();
+        head_transaction.attachment_timestamp_lower = 0;
+        head_transaction.attachment_timestamp_upper = 3812798742493;
```
bingyanglin: It's good to add explanation for this number.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
samuel-rufi: Agree Alex, sounds better and is easier to understand.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
thibault-martinez: We usually differentiate IoT and Internet though, I agree with @Alex6323, I'm not sure we should emphasise on IoT here.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+
+let outgoing_bundle = outgoing_bundle_builder
+    .push(withdrawing_tx_builder)
+    .push(depositing_tx_builder)
+    // Seal the bundle, adding indices to its transactions and assigning a bundle hash. This also prevents
+    // pushing more transactions into the bundle.
+    .seal()?
+    // Withdrawal transactions are signed, providing some signature scheme, seed, and wallet. If none are present, this
+    // step can be skipped.
+    .sign(signature_scheme, seed, wallet)?
```
samuel-rufi: Does this suggest that multiple signature schemes are allowed like WOTS/ECC/...
If so, is it really needed?

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+their veracity before constructing a `Bundle` type. The `OutgoingBundleBuilder` takes in `TransactionBuilder`s,
+and is primarily responsible for filling in all information that can only be set in the actual finalization
+of the outgoing message into a `Bundle`, such as the bundle hash.
+
+## Transaction
+
+A transaction is a sequence of 15 fields of constant length. A transaction has a total length of 8019 trits. The fields'
+name, description, and size are summarized in the table below, in their order of appearance in a transaction.
+
+| Name                              | Description                                             | Size (in trits) | Set by (functions detailed below)             |
```
thibault-martinez: Need to change the functions name in `Set by` column. And remove `add_message`, `add_input` and `add_output`.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+impl TransactionBuilder {
+    fn tag<T: TryInto<Tag>>(&mut self, tag: T) -> Result<&mut Self, TransactionBuilderError> {
+        self.tag.replace(tag.try_into()?);
+        self
+    }
+
+    /// Signs the `TransactionBuilder` by using the signature scheme `S` with some seed and input.
+    ///
+    /// TODO: This depends on one trait `SignatureScheme` which is not implemented yet.
+    fn sign<I, S>(&mut self, signature_scheme: S, seed: S::Seed, input: I) -> Result<&mut Self, TransactionBuilderError>
```
thibault-martinez: It's not possible to sign individual transactions if you consider security level > 1

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+    /// `inputs` is an object that contains all unused addresses of a seed with a sufficient balance.*
+    /// 
+    /// TODO: This function relies on some `Sponge` trait, which is not yet defined. The code below lays out
+    ///       how the sponge is thought to be used.
+    pub fn seal<S: Sponge<Output=BundleHash>>(self, sponge: S) -> Result<SealedOutgoingBundleBuilder, OutgoingBundleBuilderError> {
+        unimplemented!()
+
+        let mut current_index = 0;
+
+        for transaction_draft in bundle_builder {
```
thibault-martinez: We use both `transaction_draft` and `transaction_builder` in this RFC, can we harmonize ?

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+    /// The bundle hash is generated with a sponge by iterating through the bundle, from `0` to `last_index`, absorbing the
+    /// bundle essence of each transaction and eventually squeezing the bundle hash from the sponge.
+    ///
+    /// NOTE: `iri v1.8.1` uses `Kerl` as a sponge.
+    ///
+    /// NOTE: This function is defined on: `SealedOutgoingBundleBuilder`, `OutgoingBundleBuilder`, and `IncomingBundleBuilder`.
+    ///
+    /// TODO: This function relies on some `Sponge` trait, which is not yet defined. The code below lays out
+    ///       how the sponge is thought to be used.
+    pub fn calculate_hash<S: Sponge<Output=BundleHash>>(&self, mut sponge: S) -> S::Output {
```
thibault-martinez: Is it worth thinking of a default implementation for this ? That's quite some duplication atm.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+            transaction_draft.attachment_timestamp = timestamp();
+            transaction_draft.attachment_timestamp_lower = 0;
+            transaction_draft.attachment_timestamp_upper = 3812798742493;
+
+            if transaction_draft.tag.is_empty() {
+                transaction_draft.tag = transaction_draft.obsolete_tag;
+            }
+
+            transaction_draft.set_nonce_from_proof_of_work(proof_of_work);
+            next_local_trunk = transaction_draft.calculate_hash(sponge);
```
thibault-martinez: You are setting `next_local_trunk` but not using it.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+        // TODO: Explain what this loop is trying to achieve.
+        for transaction_draft in self {
+            if transaction_draft.value < 0 {
+                if transaction_draft.current_index >= current_index {
+                    let input = inputs[transaction_draft.address];
+
+                    // NOTE: The specific arguments to the signature scheme depend on the implementation of `SignatureScheme`.
+                    //
+                    // NOTE: The specific arguments to wallet depend on the implementation of `Wallet`.
+                    let fragments = transaction_draft.sign(signature_scheme, seed, wallet.get(transaction_draft.address));
```
thibault-martinez: It's not on the `transaction_draft` scope. Will be on private key scope from Signing Scheme.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+        }
+
+        // NOTE: We do not provide an example of `validate_withdrawal_transactions`. Its purpose is to walk
+        //       through those transactions in the bundle builder and verify their signatures.
+        bundle_builder.validate_withdrawal_transactions()?;
+
+        Ok(Self)
+    }
+
+    pub fn build(self) -> Result<Bundle, OutgoingBundleError> {
```
thibault-martinez: In the end, `validate` and `build` from `IncomingBundleBuilder ` and `SealedOutgoingBundleBuilder ` should be the same. What can we do for it ?

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+        unimplemented!()
+    }
+}
+```
+
+## How this is used
+
+In this section, we describe the algorithms needed to build a `Bundle`. The work flow depends on whether one is
```
thibault-martinez: > In this section, we describe the algorithms needed to build a `Bundle`.

Doesn't really make sense anymore.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
bingyanglin: In general, IoT consists of end devices (sensors and actuators), gateways, and the clouds (including internet). The applications of an IoT system usually include sensing, data analytics, decision making, and actions. So regarding IOTA as a data transfer layer for IoT makes sense.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Fixed for all instances of *we* in the text.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Reformulated that section.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: To me, the term IoT also includes the general internet. At the end of the day, in IoT you are also sending information from full sized computers to embedded devices and the other way round.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Fixed. That formulation was awful. :)

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: I have added a sentence about converting from `&[i8]`. In general, I'd like to keep the representation as bytes, but conversion form `&[i8]` is fine. Also, we can change this down the road if we realize that this is not nice.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Fixed and removed one getter method. The implementation phase should define more constructors.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Added a `FIXME` to the code because I am not sure where that number comes from.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: That depends on how we implement `sign`. If `sign` takes a `S: SignatureScheme`, then yeah --- it would allow for different schemes. If it takes a `WOTS`, then no. Either way, we'd probably pass the scheme in explicitly because it might need to be configured.

But let's see what its RFC brings. I have made a note in the RFC.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: I have removed the column entirely. It does not contain a lot of information at the moment, and I see this to be more relevant when describing the actual implementation.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Replaced in favour of `transaction_builder`.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: I am not totally sure about this, so I am going to leave this until after we have a `Sponge` trait.

I have made a not of this in the RFC.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Good point, I have removed the function on `TransactionBuilder` and made a note of this in the open questions.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Fixed.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: I have made a note of this in the sample code.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: Removed.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
SuperFluffy: We can probably implement one in terms of the other. I have made a note in the sample code, but would leave this to the implementation phase.

text/0000-iri-v181-transaction-and-bundle/0000-iri-v181-transaction-and-bundle.md
```diff
@@ -0,0 +1,980 @@
...
+```
+
+### `TransactionBuilder` struct
+
+The `TransactionBuilder` allows setting all the fields of a transaction, and verifies and builds a correct `Transaction`
+type. The order of the fields in `TransactionBuilder` follows the table showing the different fields of a transaction,
+the same as `Transaction`.
+
+```rust
+pub struct TransactionBuilder {
```
samuel-rufi: Just a little thing here, but the TransactionBuilder actually doesn't build the Transaction, it's the BundleBuilder which fills indices, bundle hash, etc. and finally builds it right?
So maybe we should call it `TransctionDraft` instead of `TransactionBuilder`, as you suggested before
Letting the BundleBuilder then direclty operate on the `TransctionDraft` (with all fields there: bundlehash, indices, timestamps, ... ) seems clean to me :+1:

SuperFluffy:
The team discussed the current proposal and agreed to transition it to the **pre-activated** stage.

Even though there are some blockers preventing the RFC (and by extension, its implementation) to be merged, the team agreed on the general idea behind this RFC.

In particular, the following general design points were considered to be relevant:

+ this proposal tries to only implement the transaction and bundle layout as of `iri v1.8.1`, which includes the explicit dependency on its signature, sponge/hashing, and proof of work algorithms. In particular, this proposal shall not attempt to be generic over these, because a change in either of these algorithms would constitute a breaking change with respect to the IOTA main net, thus requiring a new implementation from scratch anways;
+ this proposal is as simple as possible by only providing the ability to manually construct transactions and bundles by pushing transactions into them. This proposal does not attempt to have any more sophisticated means of constructing bundles such as determining the number of transactions needed to send a message and pushing those onto a stack.
+ this proposal tries to encode as many things as possible on the type level while still staying fairly minimal.

PR #21, opened 2019-10-14
RFC: IRI v1.8.1 cryptographic hash functions

wusyong:
[Rendered](https://github.com/wusyong/bee-rfcs/blob/crypto/text/0000-hash/0000-hash.md)

This is the draft for any hash type especially for sponge function. Other functionalities like hash algorithms, HMAC, and signature schemes should be separate RFCs. Any suggestion is welcome. The implementation would simply look like [this](https://github.com/wusyong/sponge-preivew/blob/master/src/lib.rs).

### DIFF
--- a/text/0000-hash/0000-hash.md
+++ b/text/0000-hash/0000-hash.md
@@ -5,9 +5,11 @@
 
 # Summary
 
+<!-- TODO: 81 rounds? rounds of what? -->
+
 This RFC proposes the implementation of the cryptographic hash functions `CurlP` and `Kerl`. `CurlP81`, a newtype for
 `CurlP` with 81 rounds, and `Kerl` are the two cryptographic hash functions used by `iri v1.8.1`, the IOTA reference
-implementation. Both are sponge constructions, and thus follow a very similar structure.
+implementation. Both are sponge constructions, and thus follow a very similar architecture.
 
 In `iri v1.8.1`, `CurlP81` is used for generating transaction hashes and --- through the `PearlDiver` algorithm --- for
 proof of work calculations. `Kerl` on the other hand is used as part of all other operations requiring cryptographic
@@ -45,22 +47,22 @@ For example, a transaction that withdraws funds from an IOTA address needs to ha
 - transaction hash
 - nonce
 
-The cryptographic hash functions used in `iri v1.8.1` are `Curl-P-81` and `Kerl`, which are both sponge constructions.
+The cryptographic hash functions used in `iri v1.8.1` are `CurlP81` and `Kerl`, which are both sponge constructions.
 A simplistic (and not at all complete) summary is that these are functions that are equipped with a memory state and a
 function that replaces the state memory using some input string (which can be the state memory itself). A portion of the
 memory state is then the output. In the sponge metaphor, the process of replacing the memory state by an input string is
 said to *absorb* the input, while the process of producing an output is said to *squeeze out* an output.
 
-This RFC hence proposes to implement the `Curl` and `Kerl` sponge structures, with both being implement using the same
-or very similar function signatures. The shared interface is thus defined by convention, and there is no unifying
-`Sponge` trait.
+This RFC hence proposes to implement the `CurlP` and `Kerl` sponge structures, with both using the same or very similar
+function signatures. These types thus share a common interface, which is defined by convention, instead of relying on
+a unifying `Sponge` trait.
 
 The hashes are expected to be used like this:
 
 ```rust
-// Create a Curl instance with 81 rounds.
-// This is equivalent to calling `Curl::new(81)`.
-let mut curl = Curl::default();
+// Create a CurlP instance with 81 rounds.
+// This is equivalent to calling `CurlP::new(81)`.
+let mut curlp = CurlP::default();
 
 // Assume this is a transaction and we want to digest a hash.
 let transaction = [0i8; 8019];
@@ -78,7 +80,7 @@ curl.squeeze_into(&mut tx_hash);
 let mut kerl = Kerl::default();
 
 // `Kerl::digest` is a function that combines `Kerl::absorb` and `Kerl::squeeze`. `Kerl::digest_into`
-// combines `Kerl::absorb` with `Kerl::squeeze_into`. `Curl` provides the same methods.
+// combines `Kerl::absorb` with `Kerl::squeeze_into`. `CurlP` provides the same methods.
 let tx_hash = kerl.digest(&transaction);
 ```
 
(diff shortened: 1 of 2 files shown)


thibault-martinez:
We probably need a `digest` function as well.

text/0000-hash/0000-hash.md
```diff
@@ -0,0 +1,68 @@
...
+
+This trait is going to be the interface of any sponge construction. Imagine it's like a real sponge, it `absorb`s the
+clone of input reference provided by users. And they don't need to know how internal permutation and transformation
+work, just `squeeze` the sponge and there will be a mutable reference as output. The internal state will not be cleared
+unless `reset` is called.
+
+```rust 
+pub trait Sponge
+where
+    Self: Default + Clone + Send + 'static,
```
zesterer: Are these bounds (`Send + 'static`) required for *all* sponges? If a function requires a `Send`able sponge, it can just accept `impl Sponge + Send + 'static`. This way, the requirements for `Sponge` are more relaxed.

wusyong: Thanks! I will just leave `Default + Clone` here. I think these are required at least.

SuperFluffy:
The argument for implementing a unifying `Sponge` trait for `Kerl` and `Curl` currently hinges on the assertion the following assertions:

1. code that currently uses `Curl` might drop `Kerl` in its place and start using it instead;
2. that both `Curl` and `Kerl` need to have a very similar (if not the same) API;
3. and finally, that this can only be ensured by a trait.

While using a trait indeed enforces a common interface, it is not absolutely necessary to have a similar API. We can simply choose to have a similar API by convention, and make sure that we are consistent.

The implementation of a trait should be justified by a use case where we expect a function to be generic over sponges (in this case), so that one can use the same implementation but with different arguments.

For the time being, it is not clear if such a use case exists within IOTA. Similar to the evolution of PR #20, a change in either of these algorithms would necessarily constitute a breaking change with respect to the IOTA main net.

I would thus currently not opt for a `Sponge` interface, because it does not appear to be at all required to implement a node compatible with `iri v1.8.1`. In this sense, there will not be a trait bound `S: Sponge` on any of its impl blocks or functions.

Instead, we should implement `Curl` and `Kerl`, but make sure we have a consistent interface. Other than making sure ourselves that we use common functions, we can also generate unit tests with macros that expect some function name/signature to be the same across all types.

wusyong:
Thanks for the feedbacks, I feel pretty much the same after discussion on parameterization last week and Alex's update on PoW RFC. I think it makes sense just like `Option` and `Result` usually share same method names and their logics. If you guys agree, I will change the name after pushing commits about Curl and Kerl briefing.

text/0000-hash/0000-hash.md
```diff
@@ -88,34 +88,46 @@ Useful links:
...
-provided by users. And they don't need to know how internal permutation and transformation work, just `squeeze` the
-sponge and there will be a mutable reference as output. The internal state will not be cleared unless `reset` is called. 
+The sponges proposes in this RFC have a common interface, which consists of the functions below:
+
++ `fn absorb(&mut self, input: &[Trit])`;
++ `fn squeeze(&mut self) -> Vec<Trit>`;
++ `fn squeeze_into(&mut self, output: &mut [Trit])`;
++ `fn reset(&mut self)`;
++ `fn digest(&mut self, input: &[Trit]) -> Vec<Trit>`;
++ `fn digest_into(&mut self, input: &[Trit], output: &mut [Trit])`.
```
wusyong: I really like the additional methods. It adds more flexibility for users. But we will also have to consider how long outputs are. Either a fixed size same as hash length, or we should add a parameter for output length. I would prefer a fixed size length so the output could just be an array.

thibault-martinez:
> The implementation of a trait should be justified by a use case where we expect a function to be generic over sponges (in this case), so that one can use the same implementation but with different arguments.
For the time being, it is not clear if such a use case exists within IOTA.

I want to clarify that such a use case *does* exist in IOTA and is necessary for a full compliance with version 1.8.1.

The broader use case is private networks following a specific coordinator configuration, but a more pressing issue is the distinction between Mainnet and Testnet.

text/0000-hash/0000-hash.md
```diff
@@ -0,0 +1,302 @@
...
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+<!-- TODO: 81 rounds? rounds of what? -->
+
+This RFC proposes the implementation of the cryptographic hash functions `CurlP` and `Kerl`. `CurlP81`, a newtype for
+`CurlP` with 81 rounds, and `Kerl` are the two cryptographic hash functions used by `iri v1.8.1`, the IOTA reference
+implementation. Both are sponge constructions, and thus follow a very similar architecture.
```
thibault-martinez: `iri v1.8.1` also uses CurlP27 in the testnet

text/0000-hash/0000-hash.md
```diff
@@ -0,0 +1,302 @@
...
+
+In `iri v1.8.1`, `CurlP81` is used for generating transaction hashes and --- through the `PearlDiver` algorithm --- for
+proof of work calculations. `Kerl` on the other hand is used as part of all other operations requiring cryptographic
+security, namely address generation, signature generation and verification, and bundle hash calculation.
```
thibault-martinez: If this is supposed to be exhaustive, we should mention that the coordinator also uses Kerl to produce milestones on the Mainnet.

text/0000-hash/0000-hash.md
```diff
@@ -0,0 +1,302 @@
...
+documents on [IOTA transactions and bundle].
+
+For example, a transaction that withdraws funds from an IOTA address needs to have the following set correctly:
+
+<!-- TODO: go into a bit more detail how and which hashes are used below -->
+
+- signature
+- address
+- bundle hash
+- transaction hash
```
thibault-martinez: This is implied by the nonce, a transaction hash is not `set`

text/0000-hash/0000-hash.md
thibault-martinez: I agree here, we need a size hint for methods like `fn squeeze(&mut self) -> Vec<Trit>`.

text/0000-hash/0000-hash.md
```diff
@@ -0,0 +1,302 @@
...
+        self.squeeze()
+    }
+
+    /// Digest inputs and then compute the hash with length of provided output slice
+    pub fn digest_into(&mut self, input: &[Trit], buf: &mut [Trit]) {
+        self.absorb(input);
+        self.squeeze_into(output);
+    }
+
+    pub fn new(rounds: usize) -> Self {
```
thibault-martinez: Should these be discrete values depending on a enum ?

wusyong: In the discussion on discord we mentioned const generics. But consider it will still be unstable for another year. We stay with this and wrap it with newtypes.

text/0000-hash/0000-hash.md
```diff
@@ -0,0 +1,327 @@
...
+    pub fn new() -> Self {
+        Self(CurlP::new(27))
+    }
+}
+
+struct CurlP81(CurlP);
+
+impl CurlP81 {
+    pub fn new() -> Self {
+        Self(CurlP::new(27))
```
thibault-martinez:
```suggestion
        Self(CurlP::new(81))
```

PR #24, opened 2019-10-25
RFC: `MAM Trits` module

semenov-vladyslav:


text/0000-mam-trits/0000-mam-trits.md
```diff
@@ -0,0 +1,278 @@
...
+
+```rust
+pub trait TritWord {
+    /// The number of trits per word.
+    const SIZE: usize;
+
+    /// All-zero trits word.
+    fn zero() -> Self;
+    /// Copy `n` trits from `(dx,x)` slice into `(dy,y)`.
+    fn unsafe_copy(n: usize, dx: usize, x: *const Self, dy: usize, y: *mut Self);
```
wusyong, line 40: If these methods are same as what [`unsafe`](https://doc.rust-lang.org/book/ch19-01-unsafe-rust.html) in Rust, we can make these `unsafe fn`. Probably will need to state why they are unsafe in documentation.

PR #26, opened 2019-10-28
RFC: Signing Scheme

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/rfc-signing-scheme/text/0026-signing-scheme/0026-signing-scheme.md)

Prototype: https://github.com/thibault-martinez/iota.rs/tree/master/bee-signing-scheme

### DIFF
--- a/text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
+++ b/text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
@@ -47,16 +47,18 @@ Useful links:
 
 # Detailed design
 
-The proposed traits are at the very core of asymmetric cryptography in which a pair of private and public keys are used
+The proposed design is at the very core of asymmetric cryptography in which a pair of private and public keys are used
 to create and verify digital signatures. Public keys and signatures are public while private keys remain secret.
 
-We thus provide 3 traits `PrivateKey`, `PublicKey` and `Signature`.
+## Traits
+
+Following a common asymmetric cryptography pattern, we provide 3 traits `PrivateKey`, `PublicKey` and `Signature`.
 
 Associated types are being used to bind implementations of these 3 traits together within a signing scheme. For example,
 a `PrivateKey` implementation of a signing scheme shouldn't be used with a `PublicKey` implementation of another
-signing scheme. 
+signing scheme.
 
-## `PrivateKey` trait
+### `PrivateKey` trait
 
 <!-- TODO -->
 
@@ -70,7 +72,7 @@ pub trait PrivateKey {
 }
 ```
 
-## `PublicKey` trait
+### `PublicKey` trait
 
 <!-- TODO -->
 
@@ -83,7 +85,7 @@ pub trait PublicKey {
 }
 ```
 
-## `Signature` trait
+### `Signature` trait
 
 <!-- TODO -->
 
--- a/text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
+++ b/text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
@@ -30,6 +30,13 @@ Useful links:
 
 # Detailed design
 
+The proposed traits are at the very core of asymmetric cryptography in which a pair of private and public keys are used
+to create and verify digital signatures. Public keys and signatures are public while private keys remain secret.
+
+We thus provide 3 traits `PrivateKey`, `PublicKey` and `Signature`.
+
+## `PrivateKey` trait
+
 ```rust
 pub trait PrivateKey {
     type PublicKey;
@@ -40,6 +47,8 @@ pub trait PrivateKey {
 }
 ```
 
+## `PublicKey` trait
+
 ```rust
 pub trait PublicKey {
     type Signature;
@@ -48,6 +57,8 @@ pub trait PublicKey {
     fn verify(&self, message: &[i8], signature: &Self::Signature) -> Result<(), Self::Error>;
 }
 ```
+
+## `Signature` trait
 
 ```rust
 pub trait Signature {
--- a/text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
+++ b/text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
@@ -60,6 +60,11 @@ signing scheme.
 
 ### `PrivateKey` trait
 
+The creation of a private key differs a lot from a signing scheme to another so we do not enforce a spececific way to
+to build one with a trait method. We expect each private key implementation to provide a `new` method with appropriate
+parameters. Once a private key is created, it is responsible for the creation of public keys and signatures which are
+enforced by trait methods.
+
 <!-- TODO -->
 
 ```rust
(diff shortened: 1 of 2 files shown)


text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
```diff
@@ -0,0 +1,88 @@
...
+
+## `PrivateKey` trait
+
+<!-- TODO -->
+
+```rust
+pub trait PrivateKey {
+    type PublicKey;
+    type Signature;
+
```
semenov-vladyslav: Maybe add `fn generate_private_key(rng: &RNG) -> Self` for some `RNG`? `RNG` can be implementation-specific, ie. also add `type RNG;` to this trait.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
```diff
@@ -0,0 +1,88 @@
...
+    fn generate_public_key(&self) -> Self::PublicKey;
+    fn sign(&self, message: &[i8]) -> Self::Signature;
```
semenov-vladyslav: In the most general case signature scheme signs a `message`, but in most recent designs it's just `hash`.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
```diff
@@ -0,0 +1,88 @@
...
+
+## `Signature` trait
+
+<!-- TODO -->
+
+```rust
+pub trait Signature {
+    type PublicKey;
+
+    fn recover_public_key(&self, message: &[i8]) -> Self::PublicKey;
```
semenov-vladyslav: `recover` only works for WOTS, in general for a signature scheme it can't be implemented.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
```diff
@@ -0,0 +1,88 @@
...
+    fn verify(&self, message: &[i8], signature: &Self::Signature) -> Result<(), Self::Error>;
+}
+```
+
+## `Signature` trait
+
+<!-- TODO -->
+
+```rust
+pub trait Signature {
```
semenov-vladyslav: Maybe simplify and make `Signature` just `[i8]`? The only trait it exposes is its size, a signature should only be serialized to/from trits.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
semenov-vladyslav: Should `message` be of type `Trits` from [PR#10](https://github.com/iotaledger/bee-rfcs/pull/10)?

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
```diff
@@ -0,0 +1,88 @@
...
+## `PublicKey` trait
+
+<!-- TODO -->
+
+```rust
+pub trait PublicKey {
+    type Signature;
+    type Error;
+
+    fn verify(&self, message: &[i8], signature: &Self::Signature) -> Result<(), Self::Error>;
```
semenov-vladyslav: `verify` should really return `bool`. It's hard to define the exact reason/`Error` for failure.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
```diff
@@ -0,0 +1,88 @@
...
+
+<!-- TODO -->
+
+# Rationale and alternatives
+
+<!-- TODO -->
+
+# Unresolved questions
+
+- Should we prefix `Signing` to the `PrivateKey` and `PublicKey` traits since we expect encryption keys to be developed
```
semenov-vladyslav: Better put them in separate modules `sig` for signature, `kem` for key encapsulation method.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
wusyong: It doesn't need to be associate type, just add trait bound to the parameter should do the work
`fn generate_private_key<R: Rng>(rng: &mut R) -> Self`.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
```diff
@@ -0,0 +1,88 @@
...
+
+We thus provide 3 traits `PrivateKey`, `PublicKey` and `Signature`.
+
+## `PrivateKey` trait
+
+<!-- TODO -->
+
+```rust
+pub trait PrivateKey {
+    type PublicKey;
```
wusyong: I'm not sure if trait should have associate type with each other. It can be trait bound for each other and users are not able to re-implement them because of orphan rule.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
SuperFluffy: We could create `trait RecoverablePublicKey: PublicKey` to encode this on the type level.

Are there other schemes that allow for this?

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
thibault-martinez: As discussed, private key generation is delegated to specific implementations because they are too different one from another.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
tsvisabo: so the question here is: should message be a hash or the actual bundle's message which internally can be hashed?
i think you can pass a hash as message as well so the implementation does not have to be coupled with bundle logic

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
tsvisabo: would you suggest @semenov-vladyslav  a `validate_(&self, message: &[i8]) -> bool` instead?

tsvisabo (changes requested):

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
thibault-martinez: Can you elaborate ?

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
thibault-martinez: I'll add another trait as suggested by @SuperFluffy because MSSs are based on OTS with a recover feature so we need a way to enforce it.

text/0026-signing-scheme-traits/0026-signing-scheme-traits.md
thibault-martinez: I think we'll go with the most general case by signing a message whatever it is.

thibault-martinez: Yes but following our roadmap, this will happen in v0.2

PR #27, opened 2019-11-06
RFC: `Network` module

samuel-rufi:


zesterer:
Perhaps we should discuss moving this to an `async` architecture? It seems like most of the rest of Bee's design is moving that way.

samuel-rufi:
@zesterer yes, i think we should move it to an `async` architecture.

zesterer:
Seems good so far!

text/0001-bee-network/0001-bee-network.md
```diff
@@ -0,0 +1,206 @@
...
+
+```rust
+pub trait Connection: Sized {
+
+    type InitConfig;
+    type Error;
+
+    fn connect(config: Self::InitConfig) -> Result<Self, Self::Error>;
+    fn send(&mut self, msg: Message) -> Result<(), Self::Error>;
+    fn recv(&mut self, size: u64) -> Result<Message, Self::Error>;
```
zesterer: What does this size represent?

text/0001-bee-network/0001-bee-network.md
```diff
@@ -0,0 +1,206 @@
...
+        Ok(TcpConnection{ stream })
+        
+    }
+
+    fn send(&mut self, msg: Message) -> Result<(), Self::Error> {
+       self.stream.write(msg.bytes())?;
+    }
+
+    fn recv(&mut self, size: u64) -> Result<Message, Self::Error> {
+       self.stream.read(&mut [0; size])?;
```
zesterer: Arrays are stack objects and so must have a constant length. You'll need use a dynamically-sized type like `Vec` for this.

text/0001-bee-network/0001-bee-network.md
samuel-rufi: The idea was to read a specific amout of data from the stream. Wouldn't you make it variable?

wusyong: I think it will be better if we can sort types of data into enum or just implement a trait object `dyb`. A parameter for length could always have some ways to exploit it.

text/0001-bee-network/0001-bee-network.md
```diff
@@ -159,37 +150,46 @@ impl Router {
...
+        Ok((protocol_ref_mut.send(msg)?))
 
     }
 
     pub fn recv(&self, id: PeerId) -> Result<Vec<u8>, Error> {
 
-        match self.connections.get_mut(&id).ok_or(ErrorKind::NotFound)?.deref_mut() {
-            ProtocolType::Tcp(tcp) => Ok(tcp.recv()?),
-        }
+        let mut map_ref_mut= self.connections.get_mut(&id).ok_or(ErrorKind::NotFound)?;
```
zesterer, line 180: You should be able to just use `*&*&foo` here

samuel-rufi: Do you mean `&*&self.connections.get_mut(&id).ok_or(ErrorKind::NotFound)?`? It returns `&DashMapRefMut<PeerId,Box<Protocol>>`

samuel-rufi:
This RFC describes a synchronous interface and represents the initial step towards an asynchronous solution. With the opening of https://github.com/iotaledger/bee-rfcs/pull/29 this RFC can be closed.

PR #29, opened 2020-02-05
RFC: Networking layer

samuel-rufi:
This RFC proposes the implementation of the networking layer which provides functionality to maintain neighboring nodes and exchange messages independent of the semantics of the underlying bytes. Furthermore, this RFC proposes a asynchronous solution, that aims to be faster and use fewer resources than a corresponding threaded implementation.

[Rendered](https://github.com/rufsam/bee-rfcs/blob/bee-network-async/text/0029-bee-network/0029-bee-network.md)

The prototype code can be found here: [preview-prototype](https://github.com/Alex6323/bee-p/tree/samuel-network/bee-network)
An example, with a simple receive and gossip logic that makes use of the networking layer, can be found below. It allows to peer nodes, share messages and remove nodes: [node-example](https://github.com/Alex6323/bee-p/tree/samuel-network/bee-network/examples)

### DIFF
--- a/text/0001-bee-network/0001-bee-network.md
+++ b/text/0001-bee-network/0001-bee-network.md
@@ -1,0 +1,608 @@
++ Feature name: `bee-network`
++ Start date: 2019-02-04
++ RFC PR: [iotaledger/bee-rfcs#27](https://github.com/iotaledger/bee-rfcs/pull/27)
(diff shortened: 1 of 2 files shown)


samuel-rufi:
Since this network architecture has been replaced by a more refined one, this PR can be closed.

PR #30, opened 2020-05-10
RFC: `protocol-messages`

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/protocol-messages/text/0030-protocol-messages.md)

### DIFF
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -16,6 +16,12 @@ and currently being used by [IRI](https://github.com/iotaledger/iri) nodes and
 
 # Detailed design
 
+This section details:
+- The `Message` trait that provides serialization and deserialization of messages to and from byte buffers;
+- A type-length-value protocol - on top of the trait - that adds metadata in order to send and receive the messages over
+  a transport layer;
+- The current `Message` implementations representing handshake, requests, responses, events, ...;
+
 ## `Message` trait
 
 The `Message` trait is protocol agnostic and only provides serialization and deserialization to and from byte buffers.
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -15,17 +15,14 @@ and currently being used by [IRI](https://github.com/iotaledger/iri) nodes and
 
 # Detailed design
 
-## Trait
+## `Message` trait
 
-The `Message` trait provides serialization/deserialization primitives but does not deal with protocol metadata at all
-and is not expected to be used standalone.
+The `Message` trait is protocol agnostic and only provides serialization and deserialization to and from bytes buffers.
+It should not be used as is but rather be paired with a higher layer - like a type-length-value encoding - and as such
+does not provide any safety check on inputs/outputs.
 
 ```rust
 /// A trait describing the behavior of a message.
-///
-/// This trait is protocol agnostic and only provides serialization and deserialization to and from bytes buffers.
-/// It should not be used as is but rather be paired with a higher layer - like a type-length-value encoding - and as
-/// such does not provide any safety check on inputs/outputs.
 pub(crate) trait Message {
     /// The unique identifier of the message within the protocol.
     const ID: u8;
@@ -34,13 +31,6 @@ pub(crate) trait Message {
     fn size_range() -> Range<usize>;
 
     /// Deserializes a bytes buffer into a message.
-    ///
-    /// # Arguments
-    ///
-    /// * `bytes`   -   The bytes buffer to deserialize from.
-    ///
-    /// # Panics
-    ///
     /// Panics if the provided buffer has an invalid size.
     /// The size of the buffer should be within the range returned by the `size_range` method.
     fn from_bytes(bytes: &[u8]) -> Self;
@@ -49,13 +39,6 @@ pub(crate) trait Message {
     fn size(&self) -> usize;
 
     /// Serializes a message into a bytes buffer.
-    ///
-    /// # Arguments
-    ///
-    /// * `bytes`   -   The bytes buffer to serialize into.
-    ///
-    /// # Panics
-    ///
     /// Panics if the provided buffer has an invalid size.
     /// The size of the buffer should be equal to the one returned by the `size` method.
     fn into_bytes(self, bytes: &mut [u8]);
@@ -65,7 +48,8 @@ pub(crate) trait Message {
 ## Type-length-value protocol
 
 The [type-length-value](https://en.wikipedia.org/wiki/Type-length-value) module is a safe layer on top of the messages.
-It allows serialization/deserialization to/from bytes ready to be sent/received to/from a transport layer.
+It allows serialization/deserialization to/from bytes ready to be sent/received to/from a transport layer by prepending
+or reading a header containing the type and length of the payload.
 
 ### Header
 
@@ -88,26 +72,14 @@ pub(crate) struct Tlv {}
 
 impl Tlv {
     /// Deserializes a TLV header and a bytes buffer into a message.
-    ///
-    /// # Arguments
-    ///
-    /// * `header`  -   The TLV header to deserialize from.
-    /// * `bytes`   -   The bytes buffer to deserialize from.
-    ///
-    /// # Errors
-    ///
-    /// * The advertised message type does not match the required message type.
-    /// * The advertised message length does not match the buffer length.
-    /// * The buffer length is not within the allowed size range of the required message type.
+    /// * The advertised message type should match the required message type.
+    /// * The advertised message length should match the buffer length.
+    /// * The buffer length should be within the allowed size range of the required message type.
     pub(crate) fn from_bytes<M: Message>(header: &Header, bytes: &[u8]) -> Result<M, TlvError> {
         ...
     }
 
     /// Serializes a TLV header and a message into a bytes buffer.
-    ///
-    /// # Arguments
-    ///
-    /// * `message` -   The message to serialize.
     pub(crate) fn into_bytes<M: Message>(message: M) -> Vec<u8> {
         ...
     }
@@ -237,6 +209,7 @@ partially or completely filled with `0`s. For this reason, trailing `0`s of the 
 a compression rate up to nearly 82%. Only the `payload` field is altered during this compression and the order of the
 fields stays the same.
 
+Proposed functions:
 ```rust
 pub(crate) fn compress_transaction_bytes(bytes: &[u8]) -> Vec<u8> {
     ...
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -104,14 +104,8 @@ signature in the `Message` trait. Implementations are then expected to provide a
 
 ### Endianness
 
-All multi-byte number fields of the messages of the protocol are represented as [big-endian](https://en.wikipedia.org/wiki/Endianness).
-
-### Derived traits
-
-The following traits are expected to be derived by every `Message` implementation:
-
-- `Default` which is very convenient for the implementation of `Message::from_bytes`;
-- `Clone` which is necessary to provide ownership in the context of a message broadcast;
+All multi-byte number fields of the messages of the protocol are represented as
+[big-endian](https://en.wikipedia.org/wiki/Endianness).
 
 ### Version 0
 
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -9,10 +9,9 @@ This RFC introduces the IOTA protocol messages.
 
 # Motivation
 
-Both [IRI](https://github.com/iotaledger/iri) nodes and [HORNET](https://github.com/gohornet/hornet) nodes operate on
-the IOTA networks with the TLV protocol presented in this RFC. To be able to take part in the networks, Bee nodes need
-to implement the exact same protocol; however, IRI does not implement all the messages and as such Bee nodes and IRI
-nodes will not be able to communicate.
+To be able to take part in the IOTA networks, Bee nodes need to implement the exact same protocol presented in this RFC
+and currently being used by [IRI](https://github.com/iotaledger/iri) nodes and
+[HORNET](https://github.com/gohornet/hornet) nodes.
 
 # Detailed design
 
@@ -250,7 +249,8 @@ pub(crate) fn uncompress_transaction_bytes(bytes: &[u8]) -> [u8; 1604] {
 
 # Drawbacks
 
-There is no specific drawback occurring because of the introduction of this protocol.
+Since IRI nodes only implement version `0` and `1` and Bee nodes only implement versions `0` and `2`, they will not be
+able to communicate with each other.
 
 # Rationale and alternatives
 
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -53,10 +53,12 @@ trait Message {
 ```
 
 **Notes**:
+- `size_range` returns an allowed range for the message size because some parts of some messages can be trimmed. It is
+  used to check if a message coming from a transport layer has a valid size. More details on compression below;
+- `from_bytes`/`into_bytes` panic if incorrectly used, only the following safe TLV module should directly use them;
 - `into_bytes` does not allocate a buffer because the following TLV protocol implies concatenating a header inducing
   another allocation. Since this is a hot path, a slice of an already allocated buffer for both the header and payload
   is expected; hence, limiting the amount of allocation to the bare minimum;
-- `from_bytes`/`into_bytes` panic if incorrectly used, only the following safe TLV module should directly use them;
 
 ## Type-length-value protocol
 
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -17,7 +17,7 @@ and currently being used by [IRI](https://github.com/iotaledger/iri) nodes and
 
 ## `Message` trait
 
-The `Message` trait is protocol agnostic and only provides serialization and deserialization to and from bytes buffers.
+The `Message` trait is protocol agnostic and only provides serialization and deserialization to and from byte buffers.
 It should not be used as is but rather be paired with a higher layer - like a type-length-value encoding - and as such
 does not provide any safety check on inputs/outputs.
 
@@ -30,7 +30,7 @@ pub(crate) trait Message {
     /// Returns the size range of the message as it can be compressed.
     fn size_range() -> Range<usize>;
 
-    /// Deserializes a bytes buffer into a message.
+    /// Deserializes a byte buffer into a message.
     /// Panics if the provided buffer has an invalid size.
     /// The size of the buffer should be within the range returned by the `size_range` method.
     fn from_bytes(bytes: &[u8]) -> Self;
@@ -38,7 +38,7 @@ pub(crate) trait Message {
     /// Returns the size of the message.
     fn size(&self) -> usize;
 
-    /// Serializes a message into a bytes buffer.
+    /// Serializes a message into a byte buffer.
     /// Panics if the provided buffer has an invalid size.
     /// The size of the buffer should be equal to the one returned by the `size` method.
     fn into_bytes(self, bytes: &mut [u8]);
@@ -46,7 +46,7 @@ pub(crate) trait Message {
 ```
 
 **Notes**:
-- `into_bytes` does not allocate a buffer because the following TLV protocol implies concatenating a header meaning
+- `into_bytes` does not allocate a buffer because the following TLV protocol implies concatenating a header inducing
   another allocation. Since this is a hot path, a slice of an already allocated buffer for both the header and payload
   is expected; hence, limiting the amount of allocation to the bare minimum;
 - `from_bytes`/`into_bytes` panic if incorrectly used, only the following safe TLV module should directly use them;
@@ -54,8 +54,8 @@ pub(crate) trait Message {
 ## Type-length-value protocol
 
 The [type-length-value](https://en.wikipedia.org/wiki/Type-length-value) module is a safe layer on top of the messages.
-It allows serialization/deserialization to/from bytes ready to be sent/received to/from a transport layer by prepending
-or reading a header containing the type and length of the payload.
+It allows serialization/deserialization to/from a byte buffer ready to be sent/received to/from a transport layer by
+prepending or reading a header containing the type and length of the payload.
 
 ### Header
 
@@ -77,7 +77,7 @@ pub(crate) struct Header {
 pub(crate) struct Tlv {}
 
 impl Tlv {
-    /// Deserializes a TLV header and a bytes buffer into a message.
+    /// Deserializes a TLV header and a byte buffer into a message.
     /// * The advertised message type should match the required message type.
     /// * The advertised message length should match the buffer length.
     /// * The buffer length should be within the allowed size range of the required message type.
@@ -85,7 +85,7 @@ impl Tlv {
         ...
     }
 
-    /// Serializes a TLV header and a message into a bytes buffer.
+    /// Serializes a TLV header and a message into a byte buffer.
     pub(crate) fn into_bytes<M: Message>(message: M) -> Vec<u8> {
         ...
     }
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -13,7 +13,7 @@ This RFC introduces the IOTA protocol messages that were initially added in
 To be able to take part in the IOTA networks, Bee nodes need to implement the exact same protocol presented in this RFC
 and currently being used by [IRI](https://github.com/iotaledger/iri) nodes and
 [HORNET](https://github.com/gohornet/hornet) nodes. However, it does not necessarily mean implementing the same versions
-of the protocol and a design decision - later explained - concludes that Bee nodes and IRI nodes will not be able to
+of the protocol. A design decision - later explained - concludes that Bee nodes and IRI nodes will not be able to
 communicate with each other.
 
 # Detailed design
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -72,23 +72,17 @@ struct Header {
 ### Methods
 
 ```rust
-/// Since the following methods have very common names, `from_bytes` and `into_bytes`, the sole purpose of this struct
-/// is to give them a proper namespace to avoid confusion.
-struct Tlv {}
+/// Deserializes a TLV header and a byte buffer into a message.
+/// * The advertised message type should match the required message type.
+/// * The advertised message length should match the buffer length.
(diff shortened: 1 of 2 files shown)


text/0030-protocol-messages.md
```diff
@@ -0,0 +1,243 @@
++ Feature name: `protocol-messages`
++ Start date: 2020-04-27
++ RFC PR: [iotaledger/bee-rfcs#30](https://github.com/iotaledger/bee-rfcs/pull/30)
++ Bee issue: [iotaledger/bee#70](https://github.com/iotaledger/bee/issues/70)
+
+# Summary
+
+This RFC introduces the IOTA protocol messages.
```
Alex6323: A link to some official documentation or other implementation (IRI/Hornet) would be helpful here as a reader seeing this RFC before any other documentation might ask himself here: what exactly are **the** IOTA protocol messages?

thibault-martinez: AFAIK there is no documentation, IRI doesn't really implement all of this protocol and there is no single place in HORNET where you can have an overview. Did you have something specific in mind ? I'm not really sure what we could link here 🤔

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,243 @@
...
+
+# Motivation
+
+To be able to take part in the IOTA networks, Bee nodes need to implement the exact same protocol presented in this RFC
+and currently being used by [IRI](https://github.com/iotaledger/iri) nodes and
+[HORNET](https://github.com/gohornet/hornet) nodes.
+
+# Detailed design
+
```
Alex6323, line 20: I would suggest adding an introductory sentence that gives a high-level overview on the architecture before diving deeper into the proposed traits and types described down below to realize that architecture.

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,243 @@
...
+## `Message` trait
+
+The `Message` trait is protocol agnostic and only provides serialization and deserialization to and from bytes buffers.
```
Alex6323: typo "from byte buffers" (edit: I see you do this consistently, but a **byte buffer** is usually something that can hold more than one byte, so no need to make this so explicitly clear)

thibault-martinez: It's on purpose, it's not a typo :)

thibault-martinez: serialization -> to, deserialization -> from

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,243 @@
...
+    /// Serializes a message into a bytes buffer.
+    /// Panics if the provided buffer has an invalid size.
+    /// The size of the buffer should be equal to the one returned by the `size` method.
+    fn into_bytes(self, bytes: &mut [u8]);
+}
+```
+
+**Notes**:
+- `into_bytes` does not allocate a buffer because the following TLV protocol implies concatenating a header meaning
+  another allocation. Since this is a hot path, a slice of an already allocated buffer for both the header and payload
```
Alex6323: The first sentence of that bullet point could be rephrased a bit I think. The rest of it is a bit confusing. I can provide more reasoning if requested.

thibault-martinez: Please do :)

text/0030-protocol-messages.md
samuel-rufi: Unfortunately this is still a PR. Not sure if we want to link it in this status: https://github.com/iotaledger/protocol-rfcs/pull/1

thibault-martinez: I'm basically introducing the same notions, IMO there is no point linking this one.

text/0030-protocol-messages.md
thibault-martinez: Ah you edited. Are you speaking about the 's' in bytes ?

Alex6323: I am not a native English speaker (so someone who is might jump in here), but `bytes buffers` - even if not incorrect - sounds strange to me.

text/0030-protocol-messages.md
Alex6323: If it's the only reference (and it seems properly written), and it doesn't hurt, then I would link it!!

thibault-martinez: IMO it hurts. This RFC in the current state is very confusing. It is somehow IRI centered by speaking of a `network rewrite`. It doesn't add anything that is not said in this RFC. I am more than happy to update and link it later when it has been properly reviewed.

thibault-martinez: From a quick glance there is even a big mistake in it...

text/0030-protocol-messages.md
thibault-martinez: Confirmed it's `byte buffer`, will edit, thanks!

text/0030-protocol-messages.md
Alex6323: Now that I have read it quite a few times, it's not so confusing anymore. Still I can't forget when I first read that sentence. I don't wanna be too nitpicky, so maybe we can just forget about it.

I think what disturbed me a bit was this 'meaning another allocation'. I suggest another verb here: inducing, resulting in, or something simlear ;)

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,243 @@
...
+
+The [type-length-value](https://en.wikipedia.org/wiki/Type-length-value) module is a safe layer on top of the messages.
+It allows serialization/deserialization to/from bytes ready to be sent/received to/from a transport layer by prepending
+or reading a header containing the type and length of the payload.
+
+### Header
+
+```rust
+/// A header for the type-length-value encoding.
+pub(crate) struct Header {
```
Alex6323: Let's discuss visibility modifiers. I am not sure that those belong into an RFC as those things change more often then we want. It has been the case during the prototype development that suddenly someone needed a certain trait/type to be public. In order to not having to keep the RFC in sync with the implementation all the time I would suggest not adding those modifiers to the RFC if they don't enhance the understanding of the proposed architecture.

Alex6323: The point here is to make clear how the `Header` type is defined/represented internally. As a reader of this RFC I don't care so much about visibility of this type when trying to understand the design decision.

thibault-martinez: What do you suggest ? Not putting a visibility modifier is actually choosing a visibility (public). I think I prefer going from private to public if needed rather than public to private which is a breaking change.

thibault-martinez: Also, everything in this RFC is purposely private to the crate.

Alex6323: that's wrong. If you don't say `pub` then the type/method is private. The `Header` type as well is private to the crate.

thibault-martinez: Yeah right sorry, following discussions at the same time and can't keep up. What is the real issue here ? Why do you think this is confusing ?

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,243 @@
...
+It should not be used as is but rather be paired with a higher layer - like a type-length-value encoding - and as such
+does not provide any safety check on inputs/outputs.
+
+```rust
+/// A trait describing the behavior of a message.
+pub(crate) trait Message {
+    /// The unique identifier of the message within the protocol.
+    const ID: u8;
+
+    /// Returns the size range of the message as it can be compressed.
```
zesterer, line 39: I think this method could do with some clarification: What does it mean to 'compress' in this specific context?

text/0030-protocol-messages.md
Alex6323: By writing `pub(crate)` you also make a statement, that the implementation contains several modules which have to share this type.

Alex6323: I am not saying that it's confusing. It's just not necessary to understand the RFC.

Alex6323: And that we should be careful about visibility statements of types in the RFC.

thibault-martinez: What I wanted to say is that you have to make a choice. Not having it `pub(crate)` is a choice. I am making the choice that IMO is the most obvious because it currently reflects the actual code.

thibault-martinez: How do you justify having a private type actually being used all over your crate ?

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,243 @@
...
+    /// Length of the message.
+    pub(crate) message_length: u16,
+}
+```
+
+### Methods
+
+```rust
+/// Since the following methods have very common names, `from_bytes` and `into_bytes`, the sole purpose of this struct
+/// is to give them a proper namespace to avoid confusion.
```
Alex6323: Usually namespacing is done through module definitions. Can we not have a `tlv` module and put those static methods in there?

thibault-martinez: Modules go away when you use the symbol. So there is no strict namespace.

text/0030-protocol-messages.md
Alex6323: Okay, understand. Then I would suggest changing the sentence in a way such as if Bee introduced the IOTA protocol messages for the first time.

thibault-martinez: But that's wrong ahah

text/0030-protocol-messages.md
thibault-martinez: Done

text/0030-protocol-messages.md
thibault-martinez: It's explained in the details of each message and there is a compression section at the end. But yeah maybe we can add a note.

text/0030-protocol-messages.md
Alex6323: IMHO the purpose of an RFC is not to restrict the implementor in such a way, that he can't change type visibility anymore, so leaving away the modifier (although in a Rust program is also a statement) reduces some noise to get the idea of the design across. If you want your type to keep its field private to the crate you can add `pub(crate)` to the field sure, but that doesn't give you any guarantess the implementor won't implement a `pub` function exposing this value anyway. You would have to explicitly mention that this RFC disallows such a public getter. Since this would all become too complicated I'ld suggest leaving the visibility modifiers away, and concentrate in an RFC on the type and trait modeling.

text/0030-protocol-messages.md
Alex6323: I don't get it, sorry. Why is
```rust
pub(crate) mod tlv {
   pub(crate) fn from_bytes<M: Message>(...) {}
   pub(crate) fn into_bytes<M: Message>(...) {} 
}
```
not possible?

text/0030-protocol-messages.md
thibault-martinez: But there is no "implementor", `Header` is not a trait, it's a struct. RFCs are about justifying the designs of the Bee team. Not a guide for someone to implement something. The `Header` should not leak outside of the crate. IMO removing it opens up a new kind of doubts and uncertainties. We are really bikeshedding here, there is literally no way this prevents someone from understanding the overall design.

text/0030-protocol-messages.md
thibault-martinez: Because `use tlv::from_bytes` doesn't keep `tlv`. Then you have a pretty common name in the scope which is weird. With the struct you just can't use it without `tlv::from_bytes` which is clearer.

text/0030-protocol-messages.md
Alex6323: In a way, yes, but also in a way no. Bee is a new start anyway. In 10 years maybe no one (except those involved in the early years) even knows that there was a Java implementation before Bee. (but honestly talking about **the** protocol messages without saying what it actually referes to is really really strange)

thibault-martinez: Any idea what to say then ?

text/0030-protocol-messages.md
Alex6323: Why would you import `from_bytes` from the `tlv` module? Just use `tlv::from_bytes` wherever you need it?

thibault-martinez: Why would you import anything from a module ? This is just avoiding a potential weird situation at no cost.

Alex6323: I am trying to understand this: You want to be able to write:
(wrote some gibberish Rust here)

thibault-martinez: `Tlv::from_bytes(...)` and `Tlv::into_bytes(...)` but  yes

thibault-martinez: Ah you edited. No you don't need a struct at all.

Alex6323: haha, I still don't get the issue. :joy:

Alex6323: what's so favorable about `Tlv::from_bytes(...)` over `tlv::from_bytes(...)`?

thibault-martinez: You can't get rid of the namespace.

Alex6323: but you neither

thibault-martinez: Of course you can, just importing it remove the namespace.

text/0030-protocol-messages.md
Alex6323: maybe something along the lines:
"This RFC introduces IOTA protocol messages that are described in detail further down below." ?

text/0030-protocol-messages.md
Alex6323: Unfortunately you missed the point I am trying to make. I am talking about the visibility modifier (see my initial comment).
Of course there is an implementor to an RFC, and in general the one implementing it is not identical with the person who wrote it. At least this is how I imagined the RFC process to work. I agree, that atm the situation is kind of twisted where we use RFCs after the fact to justify our implementation. But in the future those will turn into real proposals how to change/improve/extend the framework. And its a bit unfair tbh to say I made an argument this would prevent someone from understanding the overall design. Of course it does not. All I am arguing for is that visibility modifiers are not necessary to understand that design.

thibault-martinez: If we remove it, there is the immediate question of why the implementation doesn't follow the specification. If we are fine answering that whenever it's being asked maybe it's alright. I just find it very weird and still don't understand by what it's being motivated. As I said, RFCs are not definitive, they can be updated if need be, so why not having them reflect the current state of the implementation ? Making an arbitrary choice assuming it will be easier to understand is highly subjective to me.

Alex6323: well you have the same question if you change `pub(crate)` to `pub` and forget about updating the RFC, I don't see how what you do atm is any different from that.

You seem to want to make the RFCs so concrete, that we need to maintain two codebases. If we see RFCs as mere proposals that are **not** strict specifications we wouldn't have to do that.

Alex6323: Remember that you also had doc comments in the RFC. If you had changed a tiny bit in such a comment, someone could ask you about it.

thibault-martinez: No. I am just choosing the version that has the highest probably of staying like that. Header **is** `pub(crate)` and it probably will never change.

Anyway, I will remove.

Alex6323: Such a high probability needs to be found for **any** proposed type in an RFC (not just for `Header`) if we go that route. But yea, I wanted to bring this up to hear your and other opinions on that. I am not someone who can't be convinced of the opposite :smile:

thibault-martinez: Not only `Header` is `pub(crate)` in this RFC.

Alex6323: "any proposed type in an(y) RFC" :trollface:

Alex6323: I think we can close this. Thanks for the discussion, Thibault! :heart:

thibault-martinez: I think I was going to do that for all my RFCs because it's the obvious thing to do for me

Alex6323: yup, understandable. I was thinking about it for some time, until I realized that this would probably bite us at some point, because either we spend a lot of our time keeping the RFCs in sync with the code, or we'ld have "lying" RFCs. If we make absolutely clear, that RFCs only describe the *general* design decision, but aren't strict specifications, then I think the readers would still be fine with it, and we can focus on writing actual code.

Alex6323: Anyways, as you said in the call we don't need to be fully consistent on the RFCs, so I apologize if that sounded a bit like it. I simply wanted to start a discussion about how to go about it because I just didn't know, but not force anyone to change what he's most comfortable with.

thibault-martinez: Removed.

text/0030-protocol-messages.md
thibault-martinez: Removed.

text/0030-protocol-messages.md
thibault-martinez: Done

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,246 @@
...
+
+/// Serializes a TLV header and a message into a byte buffer.
+fn tlv_into_bytes<M: Message>(message: M) -> Vec<u8> {
+    ...
+}
+```
+
+## Messages
+
+Since the messages are all different one from another, there is no construction method in the `Message` trait. All the
```
Alex6323: Suggestion: "Since the various types of messages are constructed with different kind of data, there can't be a singular constructor signature in the `Message` trait."

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,246 @@
...
+|`hash`|Hash of the requested transaction.|`[u8; 49]`|49|
+
+#### `Heartbeat`
+
+Type ID: `6`
+
+A message that informs about the part of the tangle currently being fully stored by a node.
+This message is sent when a node:
+* just got paired to another node;
+* did a local snapshot and pruned away a part of the tangle;
```
Alex6323: typo: it's the **T**angle

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,246 @@
...
+It also helps other nodes to know if they can ask it a specific transaction.
+
+|Name|Description|Type|Length|
+|----|-----------|----|------|
+|`solid_milestone_index`|Index of the last solid milestone.|`u32`|4|
+|`snapshot_milestone_index`|Index of the snapshotted milestone.|`u32`|4|
+
+### Compression
+
+A transaction encoded in bytes has a length of `1604`. The `payload` field itself occupies `1312` bytes and is often
```
Alex6323: maybe too nitpicky again, but I would mention the T5B1 encoding here, something a long the lines of:
"A transaction encoded using the T5B1 codec has a length of `1604` bytes.

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,246 @@
...
+
+fn uncompress_transaction_bytes(bytes: &[u8]) -> [u8; 1604] {
+    ...
+}
+```
+
+# Drawbacks
+
+Since IRI nodes only implement version `0` and `1` and Bee nodes only implement versions `0` and `2`, they will not be
+able to communicate with each other.
```
Alex6323, line 235: Please mention, that this is only temporary until IRI implements STING.

thibault-martinez: It's not temporary. IRI will never implement STING.

samuel-rufi (dismissed):

Alex6323 (dismissed):

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,246 @@
...
++ Bee issue: [iotaledger/bee#70](https://github.com/iotaledger/bee/issues/70)
+
+# Summary
+
+This RFC introduces the IOTA protocol messages that were initially added in
+[IRI#1393](https://github.com/iotaledger/iri/pull/1393).
+
+# Motivation
+
+To be able to take part in the IOTA networks, Bee nodes need to implement the exact same protocol presented in this RFC
```
samuel-rufi, line 13: I think we should make it clear that IRI gets replaced and we focus on being compatible with Hornet. Maybe replace the Motivation part with something like this:

> To be able to take part in the IOTA networks, IOTA node softwares need to implement the exact same protocol. Since IRI is being abandoned and replaced by Hornet and Bee, Bee does not aim for compatibility with IRI. This RFC therefore addresses the protocol changes introduced by Hornet.

thibault-martinez: I feel like this is already addressed in the Drawbacks section. Maybe it's not clear enough and I can improve but I'm not sure the Motivation is the best place to say that.

samuel-rufi: Yes understand. The Drawback section notes that Bee will not be compatible with IRI, but it could be interpreted wrong when reading the Motivation part.
> [...] Bee nodes need to implement the exact same protocol presented in this RFC and currently being used by IRI nodes and HORNET nodes

thibault-martinez: It's the same protocol but not the same version though 🤔

samuel-rufi: Yes it's hard to get this right. But with the above mentioned part (https://github.com/iotaledger/bee-rfcs/pull/30#discussion_r423692148) IMO we don't have this problem and it is relatively easy to understand. But your right, not sure if it's the best place to say that.

thibault-martinez: Addressed

Alex6323 (dismissed):

samuel-rufi (dismissed):

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,248 @@
...
+- The `Message` trait that provides serialization and deserialization of messages to and from byte buffers;
+- A type-length-value protocol - on top of the trait - that adds metadata in order to send and receive the messages over
+  a transport layer;
+- The current `Message` implementations representing handshake, requests, responses, events, ...;
+
+## `Message` trait
+
+The `Message` trait is protocol agnostic and only provides serialization and deserialization to and from byte buffers.
+It should not be used as is but rather be paired with a higher layer - like a type-length-value encoding - and as such
+does not provide any safety check on inputs/outputs.
```
zesterer: In this context, what does 'safety' mean? Does it mean that Rust's safety guarantees may be invalidated with bad inputs? If so, the `from_bytes` method should probably be declared `unsafe`.

thibault-martinez: I meant that the bounds of the slice are not checked

thibault-martinez: If an invalid size is given it may panic

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,248 @@
...
+```
+
+## Messages
+
+Since the various types of messages are constructed with different kind of data, there can not be a single constructor
+signature in the `Message` trait. Implementations are then expected to provide a convenient `new` method to build them.
+
+### Endianness
+
+All multi-byte number fields of the messages of the protocol are represented as [big-endian](https://en.wikipedia.org/wiki/Endianness).
```
zesterer: Is this something that the RFC has the power to change? Most modern architectures are little-endian, and so enforcing big-endian encoding effectively mandates a significant performance hit for most machines by preventing a lot of potential compiler optimisations.

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Derived traits
+
+The following traits are expected to be derived by every `Message` implementation:
+
+- `Default` which is very convenient for the implementation of `Message::from_bytes`;
+- `Clone` which is necessary to provide ownership in the context of a message broadcast;
```
zesterer: Is broadcasting something that's expected to be supported by all message types (and can trivially be supported by all message types)? If so, a `trait Message: Clone` constraint might be beneficial such that generic code doesn't require extra trait constraints to do broadcasting.

zesterer: If this is the case, I wonder whether having `into_bytes` take `&self` is a good idea such that broadcasting can be done without cloning the contents (which might be a slow operation if the message constraints heap-allocated data structures).

text/0030-protocol-messages.md
thibault-martinez: Unfortunately no. That's the current protocol in the IOTA networks. We may have a chance to change that on an upcoming breaking change for Chrysalis though.

text/0030-protocol-messages.md
thibault-martinez: It's actually not expected to be supported by all messages. It just made it simpler to have a generic broadcast function instead of repeating it for the messages that actually do.

thibault-martinez: I like the idea of having it as a trait constraint.

thibault-martinez: Is it bad to put it as a constraint even if not used by *all* impls ? It just adds a bit of work on compile time but goes away, right ?

text/0030-protocol-messages.md
samuel-rufi: We should really keep this in mind, good suggestion.

text/0030-protocol-messages.md
thibault-martinez: Turns out I can just clone the bytes and not the entire message.

text/0030-protocol-messages.md
```diff
@@ -0,0 +1,242 @@
...
+This RFC introduces the IOTA protocol messages that were initially added in
+[IRI#1393](https://github.com/iotaledger/iri/pull/1393).
+
+# Motivation
+
+To be able to take part in the IOTA networks, Bee nodes need to implement the exact same protocol presented in this RFC
+and currently being used by [IRI](https://github.com/iotaledger/iri) nodes and
+[HORNET](https://github.com/gohornet/hornet) nodes. However, it does not necessarily mean implementing the same versions
+of the protocol and a design decision - later explained - concludes that Bee nodes and IRI nodes will not be able to
+communicate with each other.
```
Alex6323: This sentence is very confusing, It's a double negation. Not sure if on purpose.

thibault-martinez: Where is the double negation ?

zesterer (dismissed):

samuel-rufi (dismissed):

text/0030-protocol-messages.md
Alex6323: oh, I see. But in that case, there should be a comma after `of the protocol`. 

Then I suggest making it 2 sentences:
However, it does not necessarily mean implementing the same version of the protocol.
A design decision, further detailed down below, concludes, that Bee nodes and IRI nodes will not be able to communicate with each other.

thibault-martinez: Alright, will do 👍

Alex6323 (approved):
Lgtm now!

samuel-rufi (approved):

zesterer (approved):

PR #31, opened 2020-05-10
RFC: `configuration`

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/configuration/text/0031-configuration.md)

### DIFF
--- a/text/0031-configuration.md
+++ b/text/0031-configuration.md
@@ -270,10 +270,12 @@ Many configuration formats are usually considered:
 - [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md)
 - [YAML](https://yaml.org/)
 - [XML](https://www.w3.org/XML/)
+- [RON](https://github.com/ron-rs/ron)
 
-This RFC and its expected implementation are choosing `TOML` as configuration format because it is the preferred option
-in the Rust ecosystem. However, it is not excluded by RFC that other formats may be provided in the future, `serde`
-making it very easy to support other formats.
+This RFC and its expected implementations are choosing `TOML` as a first default configuration format because it is the
+preferred option in the Rust ecosystem. However, it is not excluded by this RFC that other formats may be provided in
+the future, `serde` making it very easy to support other formats. It is also important to note that TOML only provides
+a limited amount of layers of nesting due to its non-recursive syntax, which may eventually become an issue.
 
 `Serde` itself has been chosen because it is the standard for serialization/deserialization in Rust.
 
--- a/text/0031-configuration.md
+++ b/text/0031-configuration.md
@@ -143,11 +143,11 @@ let config_builder = match fs::read_to_string("config.toml") {
     Ok(toml) => match toml::from_str::<SnapshotConfigBuilder>(&toml) {
         Ok(config_builder) => config_builder,
         Err(e) => {
-            panic!("Error parsing config file: {:?}", e);
+            // Handle error
         }
     },
     Err(e) => {
-        panic!("Error reading config file: {:?}", e);
+        // Handle error
     }
 };
 
@@ -163,15 +163,15 @@ match toml::to_string(&config) {
     Ok(toml) => match fs::File::create("config.toml") {
         Ok(mut file) => {
             if let Err(e) = file.write_all(toml.as_bytes()) {
-                panic!("Error writing .toml config file: {:?}", e);
+                // Handle error
             }
         }
         Err(e) => {
-            panic!("Error creating .toml config file: {:?}", e);
+            // Handle error
         }
     },
     Err(e) => {
-        panic!("Error serializing .toml config file: {:?}", e);
+        // Handle error
     }
 }
 ```
--- a/text/0031-configuration.md
+++ b/text/0031-configuration.md
@@ -1,7 +1,7 @@
 + Feature name: `configuration`
 + Start date: 2020-05-06
-+ RFC PR: [iotaledger/bee-rfcs#00](https://github.com/iotaledger/bee-rfcs/pull/00)
-+ Bee issue: [iotaledger/bee#00](https://github.com/iotaledger/bee/issues/00)
++ RFC PR: [iotaledger/bee-rfcs#31](https://github.com/iotaledger/bee-rfcs/pull/31)
++ Bee issue: [iotaledger/bee#71](https://github.com/iotaledger/bee/issues/71)
 
 # Summary
 
@@ -253,6 +253,6 @@ making it very easy to support other formats.
 # Unresolved questions
 
 In case of binary crates, e.g.`bee-node`, configuration with CLI arguments is not described in this RFC but everything
-is already set up to support it seamlessy. The builder setters allow setting fields or overriding fields that may have
+is already set up to support it seamlessly. The builder setters allow setting fields or overriding fields that may have
 already prefilled by the parsing of a configuration file. A CLI parser library like
 [clap](https://github.com/clap-rs/clap) may be used on top of the builders.
--- a/text/0031-configuration.md
+++ b/text/0031-configuration.md
@@ -93,9 +93,12 @@ impl SnapshotConfigBuilder {
 
 A configuration type should:
 - be named `*Config`;
-- Derive `Clone` to be able to provide ownership to different components;
+- derive the following traits;
+  - `Clone` to be able to provide ownership to different components;
+  - `Serialize` if the configuration is expected to be updated and saved;
 - have the same fields with the same names, without `Option`, as the builder;
-- be immutable: no setters and no public fields;
+- have no public fields;
+- provide setters/updaters only on fields that are expected to be updatable;
 - have getters or `pub(crate)` fields;
 
 Here is a small example fitting all these requirements:
@@ -127,9 +130,7 @@ impl SnapshotConfig {
 
 ```
 
-### Read a configuration builder
-
-`toml.rs` makes it very easy to read a configuration builder from a file.
+### Read a configuration builder from a file
 
 ```rust
 let config_builder = match fs::read_to_string("config.toml") {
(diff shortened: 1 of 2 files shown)


text/0031-configuration.md
```diff
@@ -0,0 +1,280 @@
...
+        self.meta_file_path.replace(meta_file_path);
+        self
+    }
+
+    pub fn state_file_path(mut self, state_file_path: String) -> Self {
+        self.state_file_path.replace(state_file_path);
+        self
+    }
+
+    pub fn build(self) -> SnapshotConfig {
```
zesterer: I'd recommend changing this such that the initial builder method is `build` and the final creation method is `finish`, like so:

```rs
let cfg = SnapshotConfig::build()
    .meta_file_path(format!("foo"))
    .finish();
```

This more closely fits with the conventions I've seen elsewhere in the Rust ecosystem.

text/0031-configuration.md
```diff
@@ -0,0 +1,280 @@
...
+# Rationale and alternatives
+
+Many configuration formats are usually considered:
+- [TOML](https://github.com/toml-lang/toml)
+- [JSON](https://www.json.org/json-en.html)
+- [HOCON](https://github.com/lightbend/config/blob/master/HOCON.md)
+- [YAML](https://yaml.org/)
+- [XML](https://www.w3.org/XML/)
+
+This RFC and its expected implementation are choosing `TOML` as configuration format because it is the preferred option
```
zesterer: Be aware that `TOML` does somewhat restrict the data that can be serialised and cannot be used to serialise arbitrarily nested types. I'd recommend specifying that `RON` (https://github.com/ron-rs/ron) or `JSON` should be the preferred backup should `TOML` not be sufficient.

thibault-martinez: What is the limitation on nested types ? I'm actually having a section about that and it works fine ?

thibault-martinez: I can mention `RON`. I think we're just giving a default to the community but that we'll end up adapting if need be. I'll add a note!

text/0031-configuration.md
thibault-martinez: Fine by me :)

text/0031-configuration.md
```diff
@@ -0,0 +1,280 @@
...
+meta_file_path  = "./data/snapshot/mainnet.snapshot.meta"
+state_file_path = "./data/snapshot/mainnet.snapshot.state"
+```
+
+`config.rs`
+```rust
+const DEFAULT_META_FILE_PATH: &str = "./data/snapshot/mainnet.snapshot.meta";
+const DEFAULT_STATE_FILE_PATH: &str = "./data/snapshot/mainnet.snapshot.state";
+
+#[derive(Default, Deserialize)]
```
samuel-rufi, line 63: Shouldn't we return `SnapshotConfig` instead of `SnapshotConfigBuilder` when reading the configuration file from disk? Therefore move `Deserialize` to `SnapshotConfig`.

thibault-martinez: No because you may want to override everything, for example with CLI arguments

text/0031-configuration.md
zesterer: TOML only supports about 2 layers of nesting due to its non-recursive syntax.

thibault-martinez: Ahhh didn't know that! Let's see where it leads us. I'll let a comment. Switching format with serde will be very easy anyway.

zesterer (dismissed):

samuel-rufi (dismissed):

text/0031-configuration.md
```diff
@@ -0,0 +1,280 @@
...
+
+- [serde](https://github.com/serde-rs/serde): the go-to framework for **ser**ializing and **de**serializing Rust data
+  structures efficiently and generically;
+- [toml-rs](https://github.com/alexcrichton/toml-rs): a TOML encoding/decoding library for Rust with
+  serialization/deserialization on top of `serde`;
+
+## Recommendations
+
+With the following recommendations, one can:
+- read a configuration builder from a file;
```
Alex6323, line 27: read a configuration builder from file? Isn't it either "read a configuration from file", or "create a configuration builder from file"?

text/0031-configuration.md
```diff
@@ -0,0 +1,280 @@
...
+- write a configuration to a file;
+- manually override a configuration builder;
```
Alex6323: are you sure you override the configuration builder? Isn't it more like manually overriding **fields** of the configuration builder?

text/0031-configuration.md
thibault-martinez: I prefer read because that's what's really happening. You read the initial state of the builder because there is a deserialization. You never "create" it by calling `new`/`default`.

text/0031-configuration.md
thibault-martinez: Yes. I mean, overriding a field is already overriding the configuration builder you read from the file.

thibault-martinez: Want me to include fields ?

Alex6323: nah, not necessarily. If this is intentional, then I am fine with it.

text/0031-configuration.md
Alex6323: okay, just making sure, you can back it up. If you say it makes more sense, then I am fine with it.

text/0031-configuration.md
```diff
@@ -0,0 +1,287 @@
...
+- construct the configuration;
+
+### Configuration Builder
+
+The [Builder](https://rust-lang.github.io/api-guidelines/type-safety.html#c-builder) pattern is very common in Rust when
+it comes to constructing complex objects like configurations.
+
+A configuration builder type should:
+- be named `*ConfigBuilder`;
```
Alex6323: Instead of using `*` (interpretable as Deref operator or as regex 0 or many) I would either prefer `...ConfigBuilder` if you want to use a symbol, or - which I prefer most - `MyCrateConfigBuilder`.

thibault-martinez: Hmm not sure about the dots. Would "Should be suffixed with `ConfigBuilder`" be fine ?

text/0031-configuration.md
```diff
@@ -0,0 +1,287 @@
...
+            state_file_path: self.state_file_path.unwrap_or(DEFAULT_STATE_FILE_PATH.to_string()),
+        }
+    }
+}
+```
+
+### Configuration
+
+A configuration type should:
+- be named `*Config`;
```
Alex6323: *Config -> MyCrateConfig

text/0031-configuration.md
thibault-martinez: Let's not speak about crate, I already have some crates with different conf objects inside so this is confusing.

text/0031-configuration.md
```diff
@@ -0,0 +1,287 @@
...
+- derive the following traits;
+  - `Clone` to be able to provide ownership to different components;
```
Alex6323: ^ that is slightly confusing, because the way you write it one could ask: "Why is he not using Rc/Arc then?" Would be nice if you'ld explain why opting for `Clone` here ;)

text/0031-configuration.md
```diff
@@ -0,0 +1,287 @@
...
+
+    // Setters
+
+    pub fn finish(self) -> SnapshotConfig {
+        SnapshotConfig {
+            local: LocalSnapshotConfig {
+                meta_file_path: self
+                    .local
+                    .meta_file_path
+                    .unwrap_or(DEFAULT_LOCAL_SNAPSHOT_META_FILE_PATH.to_string()),
```
Alex6323, line 226: Wouldn't it be better to build those types around `str` instead of `String`. Then we wouldn't need those allocations here ^.

thibault-martinez: It can't be `str` since, when not default, it comes from the user.

text/0031-configuration.md
Alex6323: okay, yea, rephrasing should also be fine. :+1:

text/0031-configuration.md
Alex6323: alright :+1:

text/0031-configuration.md
thibault-martinez: Should I add a `unique` before ownership ? It feels weird to have a justify not providing shared ownership in this context. What would you say ?

text/0031-configuration.md
```diff
@@ -0,0 +1,287 @@
...
+```
+
+### Read a configuration builder from a file
+
+```rust
+let config_builder = match fs::read_to_string("config.toml") {
+    Ok(toml) => match toml::from_str::<SnapshotConfigBuilder>(&toml) {
+        Ok(config_builder) => config_builder,
+        Err(e) => {
+            panic!("Error parsing config file: {:?}", e);
```
Alex6323: At some point we don't want to panic here (thinking of hot-loading configs when the user makes changes to it just as in Hornet, and if he doesn't edit the file correctly, the node shouldn't implode, but instead log an error, and reject updating its parameters). So maybe you could mention that in the "Unresolved questions" section?

text/0031-configuration.md
thibault-martinez: `Since all configuration objects are most probably aggregated in a common configuration object after reading from a file, `Clone` is needed to give components a unique ownership of their configuration`.  Something like that would be fine ?

text/0031-configuration.md
thibault-martinez: This is really just to show how to use the `toml` lib, this is not intended  to be copied as is.

thibault-martinez: I can probably add a note.

Alex6323: Additionally when the node shuts down, it should write its internal parameters into the config. If it hasn't access to the file, because the user has the file open, the log should show an error that the node can't properly shutdown until it has access to the config.toml. That helps with preventing to have an unparseable config saved. It's no guarantee though. When - on startup - the node detects that the config.toml isn't parseable, then instead of panicking a log message should be printed about the error, and a proper shutdown should be initiated. So in any case, panicking is not a good option in our case.

thibault-martinez: This is very, very implementation detail IMO. This RFC is just about configuration, it shouldn't speak about the node at all.

text/0031-configuration.md
Alex6323: perfect :+1:

text/0031-configuration.md
Alex6323: I agree. I think I would still remove the explicit code in the error handler, and just put a comment there like:
```rust
// deal with parsing error
```
what do you think?

thibault-martinez: Sounds good to me 👍

Alex6323 (approved):
lgtm

zesterer (approved):

samuel-rufi (approved):

PR #33, opened 2020-05-18
RFC: `network`

Alex6323:
[Rendered](https://github.com/Alex6323/bee-rfcs/blob/bee-network/text/0033-bee-network.md)

### DIFF
--- a/text/0000-bee-network/0000-bee-network.md
+++ b/text/0000-bee-network/0000-bee-network.md
@@ -20,7 +20,7 @@ The aim of this crate is to make it simple and straightforward for developers to
 * `Peer`s: other nodes in the network (other Bees and Hornets),
 * `Message`es: the information exchanged between peers, and its serialization/deserialization.
 
-Given some endpoint identifier `epid`, sending a message to it becomes one line of asynchronous code:
+Given some identifier `epid`, sending a message to its corresponding endpoint becomes a single line of asynchronous code:
 
 ```rust
 network.send(SendMessage { epid, bytes: "hello".as_bytes() }).await?;
@@ -29,25 +29,25 @@ network.send(SendMessage { epid, bytes: "hello".as_bytes() }).await?;
 The purpose of this crate is to provide the following functionalities:
 * maintain a list of endpoints (not peers!),
 * establish and maintain connections with endpoints,
-* allow to send, multicast or broadcast byte messages of variable size to any of the endpoints,
-* allow to receive byte messages of variable size from any of the endpoints,
-* reject connections from unknown endpoints (endpoints, that are not whitelisted),
-* manage all connections asynchronously,
-* provide an extensible, yet convenient to use API.
+* allow to send, multicast or broadcast byte encoded messages of variable size to any of the endpoints,
+* allow to receive byte encoded messages of variable size from any of the endpoints,
+* reject connections from unknown, i.e. not whitelisted, endpoints,
+* manage all connections and message transfers asynchronously,
+* provide an extensible, yet convenient-to-use, and documented API.
 
 The key design decisions are now being discussed in the following sub-sections.
 
 ## async/await
 
-This crate is by nature very much dependent on events happening outside of the control of the program, e.g. listening for incoming connections from peers, waiting for packets on a specific socket, etc. Hence, this crate under the hood makes heavy use of Rust's concurrency abstractions. Luckily, with the stabilization of the `async/await` syntax, writing asynchronous code has become almost as easy to read, write, and maintain as synchronous code. In a experimental implementation of this crate the [`async_std`](https://github.com/async-rs/async-std) library was used, which comes with asynchronous drop-in replacements for their synchronous equivalents. Additionally, asynchronous mpsc channels were taken from the [`futures`](https://github.com/rust-lang/futures-rs) crate.
+This crate is by nature very much dependent on events happening outside of the control of the program, e.g. listening for incoming connections from peers, waiting for packets on a specific socket, etc. Hence, this crate under the hood makes heavy use of Rust's concurrency abstractions. Luckily, with the stabilization of the `async/await` syntax, writing asynchronous code has become almost as easy to read, write, and maintain as synchronous code. In an experimental implementation of this crate the [`async_std`](https://github.com/async-rs/async-std) library was used, which comes with asynchronous drop-in replacements for their synchronous equivalents found in Rust's standard library `std`. Additionally, asynchronous mpsc channels were taken from the [`futures`](https://github.com/rust-lang/futures-rs) crate.
 
 ## Message passing
 
-This crate favors message passing via channels over globally shared state and locks. Instead of keeping the list of endpoints in a globally accessible hashmap this crate separates and moves such state into workers that run asynchronously, and are listening for commands and events to act upon, and also notify listeners by raising their own events.
+This crate favors message passing via channels over globally shared state and locks. Instead of keeping the list of endpoints in a globally accessible hashmap this crate separates and moves such state into workers that run asynchronously, and are listening for commands and events to act upon, and also notify listeners by sending events.
 
 ## Initialization
 
-In order for this crate to be used, it has to be initialized first by passing a `NetworkConfig` to it:
+In order for this crate to be used, it has to be initialized first by providing a `NetworkConfig`:
 
 ```rust
 struct NetworkConfig {
@@ -64,40 +64,81 @@ struct NetworkConfig {
 }
 ```
 
-This has the advantage of being easily extensible and keeping the signature of the static `init` function relatively small:
+This has the advantage of being easily extensible and keeping the signature of the static `init` function small:
 
 ```rust
 fn init(config: NetworkConfig) -> (Network, Events, Shutdown) { ... }
 ```
 
-This function returns a tuple that allows the consumer of this crate to send commands (`Network`), listen to events (`Events`), and gracefully shutdown all running asynchronous tasks within the system (`Shutdown`).
+This function returns a tuple struct that allows the consumer of this crate to send commands (`Network`), listen to events (`Events`), and gracefully shutdown all running asynchronous tasks that were spawned within this system (`Shutdown`).
 
-## `Url`, `Address`, and `Port`
+## `Port`, `Address`, `Protocol`, and `Url`
 
-In order to send a message to a peer, a node needs to know the peer's address and the communication protocol it uses. This crate provides the following types to deal with that:
+In order to send a message to an endpoint, a node needs to know the endpoint's address and the communication protocol it uses. This crate provides the following abstractions to deal with that:
 
-* `Address`: a 0-cost wrapper around a `SocketAddr`:
+* `Port`: a 0-cost wrapper around a `u16`, which is introduced for type safety and better readability:
+
     ```rust
+    struct Port(pub u16);
+    ```
+
+    For convenience, `Port` dereferences to `u16`.
+
+* `Address`: a 0-cost wrapper around a `SocketAddr` which provides an adjusted, but overall simpler API than its inner type:
+
+    ```rust
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)
     struct Address(SocketAddr);
     ```
 
-    Note, that this allows to change the internal representation of an address later on.
+    An `Address` can be constructed from Ipv4 and Ipv6 addresses and a port:
 
-* `Url`: can be constructed from a `&str`. If successfull it resolves to an Ipv4 or Ipv6 address. Note, that this crate expects the Url string to always provide a scheme (e.g. `tcp://`) and a port (e.g. `15600`) when specifying a peer's address, or the implementation may either panic or return an error result.
+    ```rust
+    // Infallible construction of an `Address` from Ipv4.
+    fn from_v4_addr_and_port(address: Ipv4Addr, port: Port) -> Self { ... }
+
+    // Infallible construction of an `Address` from Ipv6.
+    fn from_v6_addr_and_port(address: Ipv6Addr, port: Port) -> Self { ... }
+
+    // Fallible construction of an `Address` from `&str`.
+    async fn from_addr_str(address: &str) -> Result<Self> { ... }
+    ```
+
+    Note, that the last function is `async`. That is because of possible delays when performing domain name resolution.
+
+* `Protocol`: an `enum`eration of supported communication protocols:
+
+    ```rust
+    #[non_exhaustive]
+    pub enum Protocol {
+        Tcp,
+    }
+    ```
+
+    Note, that future updates may support different protocols, which is why this enum is declared as `non_exhaustive`, see `Unresolved Questions` section.
+
+* `Url`: an `enum`eration that can always be constructed from an `Address` and a `Protocol` (infallible), or from a `&str`, which can fail if parsing or domain name resolution fails. If successfull however, it resolves to an Ipv4 or Ipv6 address stored in variant of the `enum` depending on the url scheme part. Note, that this crate therefore expects the Url string to always provide a scheme (e.g. `tcp://`) and a port (e.g. `15600`) when specifying an endpoint's address.
+
     ```rust
     #[non_exhaustive]
     enum Url {
-        // Represents the address of an endpoint using TCP to communicate.
+        // Represents the address of an endpoint communicating over TCP.
         Tcp(Address),
     }
     ```
 
-    Note, that future updates may support different protocols, which is why this enum is attributed as `non_exhaustive`, see `Unresolved Questions` section.
+    Note, that future updates may support different protocols, which is why this enum is declared as `non_exhaustive`, see `Unresolved Questions` section.
 
-* `Port`: a 0-cost wrapper around a `u16`, which is introduced for type safety and code readability:
     ```rust
-    struct Port(u16);
+    // Infallible construction of a `Url` from an `Address` and a `Protocol`.
+    fn new(addr: Address, proto: Protocol) -> Self { ... }
+
+    // Fallible construction of a `Url` from `&str`.
+    async fn from_url_str(url: &str) -> Result<Self> { ... }
     ```
+
+    Note, that the second function is `async`. That is because of possible delays when performing domain name resolution.
+
 
 
 ## `Endpoint` and `EndpointId`
@@ -117,7 +158,9 @@ struct Endpoint {
 }
 ```
 
-The `EndpointId` can be anything that uniquely identifies an `Endpoint`. In experiments it was implemented as a wrapper around the address.
+Note: In a peer-to-peer (p2p) network peers usually have two endpoints initially as they are actively trying to connect, but also need to accept connections from their peers. This crate is agnostic about how to handle duplicate connections as this is usually resolved by a handshaking protocol defined at a higher layer.
+
+To uniquely identify an `Endpoint`, this crate proposes the `EndpointId` type, which can be implemented as a wrapper around an `Address`.
 
 ```rust
 #[derive(Clone, Copy, Eq, PartialEq, Hash)]
@@ -155,11 +198,11 @@ enum Command {
 }
 ```
 
-The benefit of doing it this way is that the things you can do with this system is encoded on the type level, which makes it possible for the compiler to check, and therefore increases type safety. It is also very easy to add new commands which makes this crate easily extensible.
+This `enum` makes the things the consumer can do with the crate very explicit and descriptive, and also easily extensible.
 
 ### `Event`
 
-Similarly to `Command`s, the different kinds of `Event`s in the system are implemented as an `enum` allowing various types of concrete events being sent over an event channel:
+Similarly to `Command`s, the different kinds of `Event`s in the system are implemented as an `enum` allowing various types of concrete events being sent over event channels:
 
 ```rust
 enum Event {
@@ -193,7 +236,7 @@ enum Event {
 
 ```
 
-In contrast to commands though, events are messages created by the system, and those that are considered relevant are published for the user to react upon.
+In contrast to commands though, events are messages created by the system, and those that are considered relevant are published for the consumer to execute custom logic.
 
 ## Workers
 
@@ -206,9 +249,9 @@ This worker manages the list of `Endpoint`s and processes the `Command`s issued 
 
 ### TcpWorker
 
-This worker is responsible for accepting incoming TCP connection requests from other peers. Once a connection is established two additional asynchronous tasks are spawned that respectively handle incoming and outgoing messages.
+This worker is responsible for accepting incoming TCP connection requests from other endpoints. Once a connection is established two additional asynchronous tasks are spawned that respectively handle incoming and outgoing messages.
 
-Note: connection attempts from unknown IP addresses (i.e. not part of the static peer list) will be rejected.
+Note: connection attempts from unknown IP addresses (i.e. not part of the static whitelist) will be rejected.
 
 ## Shutdown
 
@@ -225,11 +268,11 @@ Apart from providing the necessary API to register channels and tasks, executing
 
 ```rust
 
-async fn execute(self) { ... }
+async fn execute(self) -> Result<()> { ... }
 
 ```
 
-This method will then send a shutdown signal to all registered workers and wait for those tasks to complete.
+This method will then try to send a shutdown signal to all registered workers and wait for those tasks to complete.
 
 # Drawbacks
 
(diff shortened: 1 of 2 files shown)


thibault-martinez (changes requested):
- Remove the intermediate folder
- Rename the file according to the PR number
- Update the header (here is the tracking issue https://github.com/iotaledger/bee/issues/76)

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
++ Feature name: `network`
++ Start date: 2020-04-29
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+This RFC proposes a networking crate (`bee-network`) to be added to the **Bee** framework that provides means to exchange byte messages with *known* endpoints.
```
thibault-martinez: Is it really necessary to mention `known` ? This won't hold for autopeering and I'm not sure to see the value in emphasizing on `known`.

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+
+# Motivation
+
+Bee nodes need a way to share messages with each other and other compatible node implementations as part of the IOTA gossip protocol, or they won't be able to form a network and synchronize with each other.
```
thibault-martinez:
```suggestion
Bee nodes need a way to share messages with each other and other compatible node implementations as part of the IOTA gossip protocol or they won't be able to form a network and synchronize with each other.
```

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+
+# Detailed design
+
+The functionalities of this crate are relatively low-level from a node perspective in the sense, that they are independent from any specifics defined by the IOTA protocol. Consequently, it has no dependencies on other crates of the framework, and can easily be reused in different contexts.
+
+The aim of this crate is to make it simple and straightforward for developers to build layers on top of it (e.g. a protocol layer) by abstracting away all the underlying networking logic. It is therefore a lot easier for them to focus on the modelling of other important aspects of a node software, like:
+
+* `Peer`s: other nodes in the network (other Bees and Hornets),
```
thibault-martinez: I'd remove `(other Bees and Hornets)` or at least add `e.g.` before. We just said it can be reused for something not IOTA-related so said like that it  sounds a bit off IMO.

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+* reject connections from unknown, i.e. not whitelisted, endpoints,
+* manage all connections and message transfers asynchronously,
+* provide an extensible, yet convenient-to-use, and documented API.
+
+The key design decisions are now being discussed in the following sub-sections.
+
+## async/await
+
+This crate is by nature very much dependent on events happening outside of the control of the program, e.g. listening for incoming connections from peers, waiting for packets on a specific socket, etc. Hence, this crate under the hood makes heavy use of Rust's concurrency abstractions. Luckily, with the stabilization of the `async/await` syntax, writing asynchronous code has become almost as easy to read, write, and maintain as synchronous code. In an experimental implementation of this crate the [`async_std`](https://github.com/async-rs/async-std) library was used, which comes with asynchronous drop-in replacements for their synchronous equivalents found in Rust's standard library `std`. Additionally, asynchronous mpsc channels were taken from the [`futures`](https://github.com/rust-lang/futures-rs) crate.
```
thibault-martinez: I would put `under the hood` at the end of the sentence, sounds weird atm.

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+
+## Message passing
+
+This crate favors message passing via channels over globally shared state and locks. Instead of keeping the list of endpoints in a globally accessible hashmap this crate separates and moves such state into workers that run asynchronously, and are listening for commands and events to act upon, and also notify listeners by sending events.
+
+## Initialization
+
+In order for this crate to be used, it has to be initialized first by providing a `NetworkConfig`:
```
thibault-martinez: Since the conf RFC has already been merged, do you think we should put a link ?

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+```rust
+struct NetworkConfig {
+    // The server or binding TCP port that remote endpoints can connect and send to.
+    binding_port: u16,
+    // The binding address that remote endpoints can connect and send to.
+    binding_addr: IpAddr,
+    // The interval between two connections attempts in case of connection losses.
+    reconnect_interval: u64,
+    // The maximum message length in terms of bytes.
+    max_message_length: usize
```
thibault-martinez: For my own curiosity, how do you envision that ? Does it split the message in chunks of this limit or does it discard it ?

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+This function returns a tuple struct that allows the consumer of this crate to send commands (`Network`), listen to events (`Events`), and gracefully shutdown all running asynchronous tasks that were spawned within this system (`Shutdown`).
+
+## `Port`, `Address`, `Protocol`, and `Url`
+
+In order to send a message to an endpoint, a node needs to know the endpoint's address and the communication protocol it uses. This crate provides the following abstractions to deal with that:
+
+* `Port`: a 0-cost wrapper around a `u16`, which is introduced for type safety and better readability:
+
+    ```rust
+    struct Port(pub u16);
```
thibault-martinez: Is that a visibility modifier I see 👀 ?

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+    ```
+
+    For convenience, `Port` dereferences to `u16`.
+
+* `Address`: a 0-cost wrapper around a `SocketAddr` which provides an adjusted, but overall simpler API than its inner type:
+
+    ```rust
+    #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)
+    struct Address(SocketAddr);
+    ```
```
thibault-martinez: For some reason, this snippet is not syntactically colored 🤔

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+    async fn from_addr_str(address: &str) -> Result<Self> { ... }
+    ```
+
+    Note, that the last function is `async`. That is because of possible delays when performing domain name resolution.
+
+* `Protocol`: an `enum`eration of supported communication protocols:
+
+    ```rust
+    #[non_exhaustive]
+    pub enum Protocol {
```
thibault-martinez: This `enum` is `pub`, the next one is not... ?

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+        Tcp,
+    }
+    ```
+
+    Note, that future updates may support different protocols, which is why this enum is declared as `non_exhaustive`, see `Unresolved Questions` section.
+
+* `Url`: an `enum`eration that can always be constructed from an `Address` and a `Protocol` (infallible), or from a `&str`, which can fail if parsing or domain name resolution fails. If successfull however, it resolves to an Ipv4 or Ipv6 address stored in variant of the `enum` depending on the url scheme part. Note, that this crate therefore expects the Url string to always provide a scheme (e.g. `tcp://`) and a port (e.g. `15600`) when specifying an endpoint's address.
```
thibault-martinez: Put Url in backquotes

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+```rust
+#[derive(Clone, Copy, Eq, PartialEq, Hash)]
+struct EndpointId(Address);
+```
+
+Note, that this `EndpointId` should be `Clone` and `Copy`, and requires to implement `Eq`, `PartialEq`, and `Hash`, because it is used as a hashmap key in several instances.
+
+## `Command`
+
+`Command`s are messages that are supposed to be issued by higher level business logic like the protocol layer. They are implemented as an `enum`, which in Rust is one way of expressing a polymorphic type:
```
thibault-martinez:
```suggestion
`Command`s are messages that are supposed to be issued by higher level business logic like a protocol layer. They are implemented as an `enum`, which in Rust is one way of expressing a polymorphic type:
```

A bit subtle but it ties this layer a bit less to *our* protocol.

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+```
+
+This `enum` makes the things the consumer can do with the crate very explicit and descriptive, and also easily extensible.
+
+### `Event`
+
+Similarly to `Command`s, the different kinds of `Event`s in the system are implemented as an `enum` allowing various types of concrete events being sent over event channels:
+
+```rust
+enum Event {
```
thibault-martinez: I think I would made this one non_exhaustive since a consumer is expected to matcher over it.

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+    TryConnect { epid: EndpointId, ... },
+}
+
+```
+
+In contrast to commands though, events are messages created by the system, and those that are considered relevant are published for the consumer to execute custom logic.
+
+## Workers
+
+There are two essential asynchronous workers in the system that are running for the node's whole lifetime. That is the `EndpointWorker` and the `TcpWorker`.
```
thibault-martinez: I'd replace `node` with `application` or whatever.

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+* `Message`es: the information exchanged between peers, and its serialization/deserialization.
+
+Given some identifier `epid`, sending a message to its corresponding endpoint becomes a single line of asynchronous code:
+
+```rust
+network.send(SendMessage { epid, bytes: "hello".as_bytes() }).await?;
+```
+
+The purpose of this crate is to provide the following functionalities:
+* maintain a list of endpoints (not peers!),
```
samuel-rufi: A reader might not understand why it's called endpoint and not peer. And if it can be considered as a p2p networking stack?  Line 8 in the "Summary" mentions endpoints, Line 20&21 mentions peers. Line 30 indicates there is a difference.

text/0000-bee-network/0000-bee-network.md
samuel-rufi: I feel like we shouldn´t mention this, even if it gives a great introducion: `Luckily, with the stabilization of the async/await syntax, writing asynchronous code has become almost as easy to read, write, and maintain as synchronous code.`

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+
+    // A connection attempt should occur.
+    TryConnect { epid: EndpointId, ... },
+}
+
+```
+
+In contrast to commands though, events are messages created by the system, and those that are considered relevant are published for the consumer to execute custom logic.
+
+## Workers
```
samuel-rufi: I think it´s important to have also some code of the workers here. The worker designs and their logic might not be implementation detail only, since it´s an essential part of this crate. Otherwise it is a bit hidden by the RFC and doing suggestions for improvement will require review of the codebase and therefore is more difficult to make.
E.g. how incoming connections are handled, or how are reading/writing sides spawned and how get messages sent to different peers / how are messages read-in might would probably cover most of it

text/0000-bee-network/0000-bee-network.md
Alex6323: oh I think it's important to make perfectly clear, that this crate requires the user to pass some endpoints he knows from somewhere, either by adding them via a config, or by running another service (autopeering), that is able to *discover* (i.e. make them known) peers/endpoints.

thibault-martinez: I see, fair enough. `known` can be a bit confusing if it's supposed to be known by the user, which is not always  the case, but sure.

text/0000-bee-network/0000-bee-network.md
Alex6323: rephrase, or complete remove?

rufsam: I would completely remove it and just move on with `In an experimental implementation [...]`

samuel-rufi: See your point, since you have the section, a brief introduction to async/await is fine - sorry. As we mentioned in the call, maybe just a simple rephrase :thumbsup:

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+## Message passing
+
+This crate favors message passing via channels over globally shared state and locks. Instead of keeping the list of endpoints in a globally accessible hashmap this crate separates and moves such state into workers that run asynchronously, and are listening for commands and events to act upon, and also notify listeners by sending events.
+
+## Initialization
+
+In order for this crate to be used, it has to be initialized first by providing a `NetworkConfig`:
+
+```rust
+struct NetworkConfig {
```
zesterer: Should `NetworkConfig` be required to implement `Default` for future compatibility?

text/0000-bee-network/0000-bee-network.md
```diff
@@ -0,0 +1,295 @@
...
+
+In order to send a message to an endpoint, a node needs to know the endpoint's address and the communication protocol it uses. This crate provides the following abstractions to deal with that:
+
+* `Port`: a 0-cost wrapper around a `u16`, which is introduced for type safety and better readability:
+
+    ```rust
+    struct Port(pub u16);
+    ```
+
+    For convenience, `Port` dereferences to `u16`.
```
zesterer: Is this API eventually intended for use with non-IP protocols?

text/0000-bee-network/0000-bee-network.md
thibault-martinez: Can you elaborate ? According to RFC 31, `NetworkConfigBuilder` has to implement `Default`, then it's just a matter of finishing it ?

zesterer:
Looks good to me, other than the comments raised above.

zesterer (dismissed):

text/0000-bee-network/0000-bee-network.md
Alex6323: hm, okay. I don't care that much about the exact words to use (do you have an idea?) as long as I get the message across, what this proposal does **not** offer. Maybe a follow-up sentence that explains what I mean by `known` would help :slightly_smiling_face:

zesterer (dismissed):

text/0000-bee-network/0000-bee-network.md
Alex6323: was actually thinking the same

text/0000-bee-network/0000-bee-network.md
Alex6323: Removed.

text/0000-bee-network/0000-bee-network.md
Alex6323: I put it at the end first, but ended up moving it right after the `hence`. I think it rings better with the flow of the argument.

text/0000-bee-network/0000-bee-network.md
Alex6323: Done.

zesterer (dismissed):

text/0000-bee-network/0000-bee-network.md
Alex6323: Ooooh, you got me there! :trollface:

Alex6323: Removed.

text/0000-bee-network/0000-bee-network.md
Alex6323: backticks :trollface: 
Done.

text/0000-bee-network/0000-bee-network.md
Alex6323: Changed.

text/0000-bee-network/0000-bee-network.md
Alex6323: ah, very good point and foresight actually. :+1:

Alex6323: done

text/0000-bee-network/0000-bee-network.md
Alex6323: Agree. Done.

text/0000-bee-network/0000-bee-network.md
Alex6323: Well, this crate is only the lowest level in the stack. At this level there are no peers yet, only TCP/UDP connections. Those connections are made up of 2 endpoints belonging to 2 peers. The whole networking stack is much bigger (protocol, peering, gossip even?), so there is that. I see your point, though. Maybe I should make this more clear, and add a `Terminology` section? What do you, and the others think?

text/0000-bee-network/0000-bee-network.md
Alex6323: yea, no worries. Maybe it's just the lax wording I use there, that bothers you a bit unconciously. At the moment I don't know how to rephrase, or what we would gain from it :slightly_smiling_face:

text/0000-bee-network/0000-bee-network.md
Alex6323: Yea, I was thinking about doing that. I could put more code into the RFC, but I think implementing the workers is pretty much straight forward and I can't think of many different ways of implementing such a worker. But I'll have a look, and see what I can add! :+1:

text/0000-bee-network/0000-bee-network.md
Alex6323: Good question! I would very cautiously assume yes, but I am totally unsure about it.

text/0000-bee-network/0000-bee-network.md
Alex6323: @zesterer yes, please elaborate? @thibault-martinez  `NetworkConfig` is supposed to be created through the builder only, right?

text/0000-bee-network/0000-bee-network.md
Alex6323: @thibault-martinez okay, I was thinking a bit about it, and came to conclusion, that maybe I shouldn't overthink it, and put too much effort into this.

What do you think of `with connected endpoints`? It kind of implies what I wanted to express.

zesterer (dismissed):

text/0000-bee-network/0000-bee-network.md
zesterer: Ah, fair enough. All good then :)

text/0000-bee-network/0000-bee-network.md
zesterer: Right. In that case, perhaps these types should go into some sort of `ip` module?

text/0000-bee-network/0000-bee-network.md
thibault-martinez: Sounds good to me 👍

text/0000-bee-network/0000-bee-network.md
thibault-martinez: Seems like it's still there :)

text/0000-bee-network/0000-bee-network.md
Alex6323: agree, we should do that at some point :+1:

Alex6323: I mean, they are already in their own module, it's just not called `ip`, but `address` instead.

text/0000-bee-network/0000-bee-network.md
Alex6323: Done.

text/0000-bee-network/0000-bee-network.md
Alex6323: no it's not (see rendered version). Somehow Gh is not able to track my changes in my `bee-network` branch anymore.

thibault-martinez: Well when I reviewed, the commit that changed it wasn't shown yet... First syndromes of bloated PR :)

text/0000-bee-network/0000-bee-network.md
samuel-rufi: It seems bit subjective. I think we should avoid things like that when possible. maybe we'll come up with something later, it's not urgent 🙂

text/0000-bee-network/0000-bee-network.md
samuel-rufi: Perfect, thanks!

text/0000-bee-network/0000-bee-network.md
samuel-rufi: Maybe a short `Terminology` section would be helpful, I think this will be also helpful for us.

text/0000-bee-network/0000-bee-network.md
Alex6323: oh, I see now that the term might be confusing. It's just the buffer size where received bytes will be stored in. We this to be allocated only once. So in the current protocol this number has to be set to 1604, which is the worst case size a message that would fill the buffer completely.

text/0000-bee-network/0000-bee-network.md
Alex6323: no, I see none :joy: jk
Thanks!

text/0000-bee-network/0000-bee-network.md
thibault-martinez: Alright but shouldn't it be a constant then ? If it's a conf we can only assume it *will* be modified. What happens if it's too small then ?

Alex6323: I am not sure what happens internally if a message with more than 1604 bytes is sent. I would assume a resize (and therefore reallocation), but could also just be a discard of the extant bytes.

text/0000-bee-network/0000-bee-network.md
Alex6323: haha, the closing bracket `]` is missing. Thanks for pointing this out. :smile:

thibault-martinez: Ah right, didn't even see it

Alex6323: Solved.

text/0000-bee-network/0000-bee-network.md
Alex6323: `Alright but shouldn't it be a constant then ?` I would not make it constant, because then people who'ld like to use it would have to edit and recompile that crate. I think it's fair to assume, that if someone wants to use this crate for his purposes he'ld like to be able to change the buffer size depending on his/her usecase.

thibault-martinez: Fair enough. As soon as we're sure of the behaviours on an insufficient buffer I'm fine.

text/0000-bee-network/0000-bee-network.md
thibault-martinez: I wouldn't mind a terminology section

text/0000-bee-network/0000-bee-network.md
thibault-martinez: I'm not entirely convinced of putting some code here 🤔 
Sounds like a lot of implementation details.
At the very least maybe only the main loop but I'm not sure if that helps...

thibault-martinez (dismissed):

zesterer (dismissed):

text/0000-bee-network/0000-bee-network.md
Alex6323: I've reread that section, and I think, it should be fine. From the context it should be clear that there is a difference:
"It is therefore a lot easier for them to focus on the modelling of other important aspects of a node software, like:
* Peer
* Message
"
I am talking there about higher layers introducing those concepts, so I don't see why that's an issue.
I think I'll just remove the *(not peers!)*, and close this discussion.
If there is regular confusion about it, I'll change the RFC text :)

text/0000-bee-network/0000-bee-network.md
Alex6323: let's not put code there for now. I can make updates to the RFC later on, if need be. :smile:

thibault-martinez (approved):

zesterer (approved):

samuel-rufi (approved):

PR #34, opened 2020-05-25
RFC: `ternary-hash`

thibault-martinez:
[Rendered](https://github.com/thibault-martinez/bee-rfcs/blob/rfc-sponges/text/0034-ternary-hash.md)

### DIFF
--- a/text/0034-ternary-hash.md
+++ b/text/0034-ternary-hash.md
@@ -6,7 +6,7 @@
 # Summary
 
 This RFC proposes the ternary `Hash` type, the `Sponge` trait, and two cryptographic hash functions `CurlP` and `Kerl`
-implementing it. The 3 cryptographic hash functions used in the IOTA current networks (i.e. as of IOTA Reference
+implementing it. The 3 cryptographic hash functions used in the current IOTA networks (i.e. as of IOTA Reference
 Implementation `iri v1.8.1`) are `Kerl`, `CurlP27`, and `CurlP81`.
 
 Useful links:
@@ -16,7 +16,7 @@ Useful links:
 + [Curl-p](https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/crypto/Curl.java)
 + [Kerl specification](https://github.com/iotaledger/kerl/blob/master/IOTA-Kerl-spec.md)
 + [`iri v1.8.1`](https://github.com/iotaledger/iri/releases/tag/v1.8.1-RELEASE)
-+ [IOTA transactions and bundle](https://docs.iota.org/docs/dev-essentials/0.1/concepts/bundles-and-transactions)
++ [IOTA transactions](https://docs.iota.org/docs/getting-started/1.0/understanding-iota/transactions)
 
 # Motivation
 
@@ -32,7 +32,8 @@ converts its binary output back to ternary. For `CurlP` specifically, its varian
 
 ## Hash
 
-This RFC defines a ternary type `Hash`. The exact definition is an implementation detail but an example definition could
+This RFC defines a ternary type `Hash` which is the base input and output of the `Sponge`.
+The exact definition is an implementation detail but an example definition could
 simply be the following:
 
 ```rust
@@ -47,45 +48,47 @@ const HASH_LENGTH: usize = 243;
 ## Sponges
 
 `CurlP` and `Kerl` are cryptographic sponge constructions. They are equipped with a memory state and a function that
-replaces the state memory using some input string (which can be the state memory itself). A portion of the memory state
-is then the output. In the sponge metaphor, the process of replacing the memory state by an input string is said to
-`absorb` the input, while the process of producing an output is said to `squeeze` out an output.
+replaces the state memory using some input string. A portion of the memory state is then the output. In the sponge
+metaphor, the process of replacing the memory state by an input string is said to `absorb` the input, while the process
+of producing an output is said to `squeeze` out an output.
 
 The hash functions are expected to be used like this:
 
-```rust
-// Create a CurlP instance with 81 rounds.
-// This is equivalent to calling `CurlP::new(CurlPRounds::Rounds81)`.
-let mut curlp = CurlP81::new();
+- `curlp`
+  ```rust
+  // Create a CurlP instance with 81 rounds.
+  // This is equivalent to calling `CurlP::new(CurlPRounds::Rounds81)`.
+  let mut curlp = CurlP81::new();
 
-// Assume we have some input that is a `Trits` slice as defined by the `bee-ternary` crate, and an output buffer
-/// `TritBuf`:
-let transaction: &Trits<T1B1>;
-let mut tx_hash: TritBuf<T1B1Buf>;
+  // Assume we have some transaction trits, all zeroes for the sake of this example.
+  let transaction = TritBuf::<T1B1Buf>::zeros(6561);
+  let mut hash = TritBuf::<T1B1Buf>::zeros(243);
 
-// Absorb the transaction.
-curlp.absorb(&Trits::from_i8_unchecked(&transaction));
-// And squeeze into the `tx_hash` buffer.
-curlp.squeeze_into(&mut tx_hash);
-```
-```rust
-// Create a Kerl instance. Kerl does not come with a configurable number of rounds.
-let mut kerl = Kerl::new();
+  // Absorb the transaction.
+  curlp.absorb(&transaction);
+  // Squeeze out a hash.
+  curlp.squeeze_into(&mut hash);
+  ```
 
-// `Kerl::digest` is a function that combines `Kerl::absorb` and `Kerl::squeeze`.
-// `Kerl::digest_into` combines `Kerl::absorb` with `Kerl::squeeze_into`. `CurlP` provides the same methods.
-let tx_hash = kerl.digest(&transaction);
-```
+- `kerl`
+  ```rust
+  // Create a Kerl instance.
+  let mut kerl = Kerl::new();
+
+  // `Kerl::digest` is a function that combines `Kerl::absorb` and `Kerl::squeeze`.
+  // `Kerl::digest_into` combines `Kerl::absorb` with `Kerl::squeeze_into`.
+  let hash = kerl.digest(&transaction);
+  ```
 
 The main proposal of this RFC are the `Sponge` trait and the `CurlP` and `Kerl` types that are implementing it.
-This RFC relies on the the presence of the types `TritBuf` and `Trits`, as defined by
+This RFC relies on the presence of the types `TritBuf` and `Trits`, as defined by
 [RFC36](https://github.com/iotaledger/bee-rfcs/blob/master/text/0036-ternary.md), which are assumed to be owning and
 borrowing collections of binary-encoded ternary in the `T1B1` encoding (one trit per byte).
 
 ```rust
 /// The common interface of cryptographic hash functions that follow the sponge construction and that act on ternary.
 trait Sponge {
-    /// An error indicating that a failure has occured during `absorb`.
+    /// An error indicating that a failure has occured during a sponge operation.
     type Error;
 
     /// Absorb `input` into the sponge.
@@ -97,7 +100,7 @@ trait Sponge {
     /// Squeeze the sponge into a buffer
     fn squeeze_into(&mut self, buf: &mut Trits) -> Result<(), Self::Error>;
 
-    /// Convenience function using `Sponge::squeeze_into` to to return an owned version of the hash.
+    /// Convenience function using `Sponge::squeeze_into` to return an owned version of the hash.
     fn squeeze(&mut self) -> Result<TritBuf, Self::Error> {
         let mut output = TritBuf::zeros(HASH_LENGTH);
         self.squeeze_into(&mut output)?;
@@ -113,7 +116,7 @@ trait Sponge {
     }
 
     /// Convenience function to absorb `input`, squeeze the sponge, and reset the sponge in one go.
-    /// Returns an owned versin of the hash.
+    /// Returns an owned version of the hash.
     fn digest(&mut self, input: &Trits) -> Result<TritBuf, Self::Error> {
         self.absorb(input)?;
         let output = self.squeeze()?;
@@ -123,33 +126,32 @@ trait Sponge {
 }
 ```
 
-Following the sponge metaphor, an input reference provided by the user is `absorb`ed, and an output will be `squeeze`d
-from the data structure. `digest` is a convenience method calling `absorb` and `squeeze` in one go. The `*_into`
-versions of these methods are for passing a buffer into which the calculated hashes are written.  The internal state
-will not be cleared unless `reset` is called.
+Following the sponge metaphor, an input provided by the user is `absorb`ed, and an output will be `squeeze`d from the
+data structure. `digest` is a convenience method calling `absorb` and `squeeze` in one go. The `*_into` versions of
+these methods are for passing a buffer into which the calculated hashes are written. The internal state will not be
+cleared unless `reset` is called.
 
 ### Design of `CurlP`
 
 `CurlP` is designed as a hash function that acts on a `T1B1` binary-encoded ternary buffer, with a hash length of `243`
-trits and an inner state of `729` trits, which we take as constants:
+trits and an inner state of `729`:
 
 ```rust
-const HASH_LENGTH: usize = 243;
 const STATE_LENGTH: usize = HASH_LENGTH * 3;
 ```
 
 In addition, a lookup table is used as part of the absorption step.
-2 is used to pad the table since there are only 9 combinations should be taken:
+
+`2` is used to pad the table since there are only 9 combinations should be taken:
 
 ```rust
 const TRUTH_TABLE: [i8; 11] = [1, 0, -1, 2, 1, -1, 0, 2, -1, 1, 0];
 ```
 
-The way `CurlP` is defined, it can not actually fail, because the input or outputs can be of arbitrary size.
-Thus, the associated type `Error = !`, or `Error = Infallible` as `!` is not yet stabilized in Rust as of version
-`1.43`.
+The way `CurlP` is defined, it can not actually fail, because the input or outputs can be of arbitrary size; hence,
+the associated type `Error = Infallible`.
 
-This `enum` defines the number of rounds of hashing to apply before a hash is squeezed.
+`CurlP` has two common variants depending on the number of rounds of hashing to apply before a hash is squeezed.
 ```rust
 #[derive(Copy, Clone)]
 pub enum CurlPRounds {
@@ -168,12 +170,12 @@ struct CurlP {
     /// The internal state.
     state: TritBuf,
 
-    /// Workspace for performing transformations
+    /// Workspace for performing transformations.
     work_state: TritBuf,
 }
 ```
 
-In addition, there are two wrapper types for the very common `CurlP` variants with `27` and `81` rounds:
+In addition, the two wrapper types for the very common `CurlP` variants with `27` and `81` rounds:
 
 ```rust
 struct CurlP27(CurlP);
@@ -193,7 +195,7 @@ impl CurlP81 {
 }
 ```
 
-For convenience, they should both dereference to an actual `CurlP``.
+For convenience, they should both dereference to an actual `CurlP`:
 
 ```rust
 impl Deref for CurlP27 {
@@ -225,7 +227,7 @@ impl DerefMut for CurlP81 {
 }
 ```
 
-This allows using them as a `Sponge` as well if there is a blanket implementation.
+This allows using them as a `Sponge` as well if there is a blanket implementation like this:
 
 ```rust
 impl<T: Sponge, U: DerefMut<Target = T>> Sponge for U {
--- a/text/0034-ternary-hash.md
+++ b/text/0034-ternary-hash.md
@@ -140,9 +140,16 @@ trits and an inner state of `729`:
 const STATE_LENGTH: usize = HASH_LENGTH * 3;
 ```
 
-In addition, a lookup table is used as part of the absorption step.
+In addition, a lookup table is used as part of the absorption step. The formula `t + (t' << 2) + 5`, where `t` and `t'`
+are two balanced trits, gives indexes to use to find the resulting trit:
 
-`2` is used to pad the table since there are only 9 combinations should be taken:
+||-1|0|1|
+|-|-|-|-|
+|**-1**|0|4|8|
+|**0**|1|5|9|
+|**1**|2|6|10
+
+Since the indexes `3` and `7` are not produced by this truth table, `2` is used to pad the table:
 
 ```rust
 const TRUTH_TABLE: [i8; 11] = [1, 0, -1, 2, 1, -1, 0, 2, -1, 1, 0];
--- a/text/0034-ternary-hash.md
+++ b/text/0034-ternary-hash.md
@@ -7,7 +7,7 @@
 
 This RFC proposes the ternary `Hash` type, the `Sponge` trait, and two cryptographic hash functions `CurlP` and `Kerl`
 implementing it. The 3 cryptographic hash functions used in the current IOTA networks (i.e. as of IOTA Reference
-Implementation `iri v1.8.1`) are `Kerl`, `CurlP27`, and `CurlP81`.
+Implementation `iri v1.8.6`) are `Kerl`, `CurlP27`, and `CurlP81`.
 
 Useful links:
 
@@ -15,7 +15,7 @@ Useful links:
 + [The sponge and duplex constructions](https://keccak.team/sponge_duplex.html)
 + [Curl-p](https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/crypto/Curl.java)
 + [Kerl specification](https://github.com/iotaledger/kerl/blob/master/IOTA-Kerl-spec.md)
-+ [`iri v1.8.1`](https://github.com/iotaledger/iri/releases/tag/v1.8.1-RELEASE)
++ [`iri v1.8.6`](https://github.com/iotaledger/iri/releases/tag/v1.8.6-RELEASE)
 + [IOTA transactions](https://docs.iota.org/docs/getting-started/1.0/understanding-iota/transactions)
 
 # Motivation
@@ -97,7 +97,7 @@ trait Sponge {
     /// Reset the inner state of the sponge.
     fn reset(&mut self);
 
-    /// Squeeze the sponge into a buffer
+    /// Squeeze the sponge into a buffer.
     fn squeeze_into(&mut self, buf: &mut Trits) -> Result<(), Self::Error>;
 
     /// Convenience function using `Sponge::squeeze_into` to return an owned version of the hash.
@@ -107,7 +107,7 @@ trait Sponge {
         Ok(output)
     }
 
-    /// Convenience function to absorb `input`, squeeze the sponge into a buffer, and reset the sponge in one go.
+    /// Convenience function to absorb `input`, squeeze the sponge into a buffer `buf`, and reset the sponge in one go.
     fn digest_into(&mut self, input: &Trits, buf: &mut Trits) -> Result<(), Self::Error> {
         self.absorb(input)?;
         self.squeeze_into(buf)?;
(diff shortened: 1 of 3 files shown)


text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+it. This RFC relies on the the presence of the types `TritBuf` and `Trits`, which are assumed to be owning and borrowing
+collections of binary-coded ternary in the `T1B1` encoding (one trit per byte), similar to
+[PathBuf](https://doc.rust-lang.org/std/path/struct.PathBuf.html) and
+[Path](https://doc.rust-lang.org/std/path/struct.Path.html).
+
+```rust
+/// The common interface of cryptographic hash functions that follow the sponge construction and that act on ternary.
+trait Sponge {
+    /// The expected length of the input to the sponge.
+    const IN_LEN: usize;
```
zesterer: Should these always be constant?

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+`CurlP` and `Kerl` are cryptographic sponge constructions. They are equipped with a memory state and a function that
+replaces the state memory using some input string (which can be the state memory itself). A portion of the memory state
+is then the output. In the sponge metaphor, the process of replacing the memory state by an input string is said to
+`absorb` the input, while the process of producing an output is said to `squeeze` out an output.
+
+The hash functions are expected to be used like this:
+
+```rust
+// Create a CurlP instance with 81 rounds.
+// This is equivalent to calling `CurlP::new(81)`.
```
zesterer: > This is equivalent to calling `CurlP::new(81)`

This is inconsistent with documentation below. Should it not be `CurlP::new(CurlPRounds::Rounds81)`?

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+
+    /// Workspace for performing transformations
+    work_state: TritBuf,
+}
+```
+
+In addition, there are two wrapper types for the very common `CurlP` variants with `27` and `81` rounds:
+
+```rust
+struct CurlP27(CurlP);
```
zesterer, line 195: Do these `impl Deref<Target=CurlP>`?

text/0034-ternary-hash.md
thibault-martinez: Good catch, the enum was added later.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+```
+
+The way `CurlP` is defined, it can not actually fail, because the input or outputs can be of arbitrary size.
+Thus, the associated type `Error = !`, or `Error = Infallible` as `!` is not yet stabilized in Rust as of version
+`1.43`.
+
+This `enum` defines the number of rounds of hashing to apply before a hash is squeezed.
+```rust
+#[derive(Copy, Clone)]
+pub enum CurlPRounds {
```
zesterer: Why not just accept a runtime constant instead of this enum? The compiler will be able to do constant inlining anyway so it shouldn't represent a performance problem.

thibault-martinez: Because Curl is only defined for 27 and 81.

text/0034-ternary-hash.md
thibault-martinez: Seems like they don't. They impl `Sponge` and forward the calls to the inner `CurlP`.

text/0034-ternary-hash.md
thibault-martinez: I assume these associated constants were added for flexibility but at the moment they are all defined to `HASH_LEN` and we surely won't have any more ternary sponge. I guess we can remove them.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+
+# Detailed design
+
+## Hash
+
+This RFC defines a ternary type `Hash`. The exact definition is implementation detail but an example definition could
+simply be the following.
+
+```rust
+struct Hash([i8; 243]);
```
samuel-rufi: Why is Hash defined as `[i8; 243]` and transaction fields like Address like this: `Address(pub(crate) TritBuf<T1B1Buf>)`?

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+
++ [The sponge and duplex constructions](https://keccak.team/sponge_duplex.html)
++ [Curl-p](https://github.com/iotaledger/iri/blob/dev/src/main/java/com/iota/iri/crypto/Curl.java)
++ [Kerl specification](https://github.com/iotaledger/kerl/blob/master/IOTA-Kerl-spec.md)
++ [`iri v1.8.1`](https://github.com/iotaledger/iri/releases/tag/v1.8.1-RELEASE)
++ [IOTA transactions and bundle](https://docs.iota.org/docs/dev-essentials/0.1/concepts/bundles-and-transactions)
+
+# Motivation
+
+In order to participate in the IOTA network, an application needs be able to construct valid messages that can be
```
Alex6323: needs **to** be able to

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+verified by nodes in the network. Conversely, a node needs to be able to verify other nodes messages. Among other
+operations, this is accomplished by checking transaction signatures (from which addresses are calculated), transaction
```
Alex6323: *from which addresses are calculated* I assume this is about "recover public key" from signatures? I think we could be a more clear here.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+hashes, and hashes of collections of transactions (so-called bundles).
```
Alex6323: .... and hashes of bundles of transactions (so-called bundle hashes).

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+
+The two hash functions currently used are both sponge constructions: `CurlP`, which is specified entirely in balanced
```
Alex6323, line 27: *...are both sponge constructions*: maybe a link to the Wikipedia article would be helpful?

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+ternary, and `Kerl`, which first converts ternary input to a binary representation, applies `keccak-384` to it, and then
+converts its binary output back to ternary. For `CurlP` specifically, its variants `CurlP27` and `CurlP81` are used.
+
+# Detailed design
+
+## Hash
+
+This RFC defines a ternary type `Hash`. The exact definition is implementation detail but an example definition could
```
Alex6323: ...is **an** implementation detail

text/0034-ternary-hash.md
thibault-martinez: There is a link in the summary already.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+simply be the following.
```
Alex6323: ...simply be the following: **colon**

thibault-martinez: There is a weird new line also ahah

text/0034-ternary-hash.md
Alex6323: alright, was just thinking that Wikipedia would be better suited for the layman, who's just trying to get the gist of it.

thibault-martinez: Sure, will add it with the other links as well.

text/0034-ternary-hash.md
samuel-rufi: Maybe this sounds better:
`In order to participate in the IOTA network, an application needs be able to construct valid messages that can be verified by other participants in the network.`
And do we need this sentence: `Conversely, a node needs to be able to verify other nodes messages.`
Seems like we are telling the same thing twice.
Also the `Conversely` might be confusing, since it compares with the first sentence.

text/0034-ternary-hash.md
thibault-martinez: I think for the sake of this RFC we don't really care since this is implementation detail.

But for the actual answer I invoke @zesterer

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+`absorb` the input, while the process of producing an output is said to `squeeze` out an output.
+
+The hash functions are expected to be used like this:
+
+```rust
+// Create a CurlP instance with 81 rounds.
+// This is equivalent to calling `CurlP::new(81)`.
+let mut curlp = CurlP81::new();
+
+// Assume we have some input that is a Trits slice as defined by the bee_ternary crate, and an output buffer `TritBuf`:
```
Alex6323: bee-ternary crate (dash instead of underscore)

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+```rust
+// Create a Kerl instance. Kerl does not come with a configurable number of rounds.
+let mut kerl = Kerl::new();
+
+// `Kerl::digest` is a function that combines `Kerl::absorb` and `Kerl::squeeze`.
+// `Kerl::digest_into` combines `Kerl::absorb` with `Kerl::squeeze_into`. `CurlP` provides the same methods.
+let tx_hash = kerl.digest(&transaction);
+```
+
+The main proposal of this RFC are the `Sponge` trait and the `CurlP` and `Kerl` types that are implemented in terms of
```
Alex6323: I think I'ld prefer: "... and the `CurlP` and `Kerl` types that are implementing it"

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+it. This RFC relies on the the presence of the types `TritBuf` and `Trits`, which are assumed to be owning and borrowing
```
Alex6323: link to the ternary RFC

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+```rust
+/// The common interface of cryptographic hash functions that follow the sponge construction and that act on ternary.
+trait Sponge {
+    /// The expected length of the input to the sponge.
+    const IN_LEN: usize;
+
+    /// The length of the hash squeezed from the sponge.
+    const OUT_LEN: usize;
+
+    /// An error indicating a that a failure has occured during `absorb`.
```
Alex6323: one **a** too much

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+
+The main proposal of this RFC are the `Sponge` trait and the `CurlP` and `Kerl` types that are implemented in terms of
+it. This RFC relies on the the presence of the types `TritBuf` and `Trits`, which are assumed to be owning and borrowing
+collections of binary-coded ternary in the `T1B1` encoding (one trit per byte), similar to
+[PathBuf](https://doc.rust-lang.org/std/path/struct.PathBuf.html) and
+[Path](https://doc.rust-lang.org/std/path/struct.Path.html).
+
+```rust
+/// The common interface of cryptographic hash functions that follow the sponge construction and that act on ternary.
+trait Sponge {
```
Alex6323, line 90: Just a thought:
Would it not have been better to split the trait into the base `Sponge` trait (absorb, squeeze, reset), and an extension trait called for example `SpongeInplace` or simply `SpongeExt`, that contains the "..._into" functions?

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+```
+
+Following the sponge metaphor, an input reference provided by the user is `absorb`ed, and an output will be `squeeze`d
+from the data structure. `digest` is a convenience method calling `absorb` and `squeeze` in one go. The `*_into`
+versions of these methods are for passing a buffer into which the calculated hashes are written.  The internal state
+will not be cleared unless `reset` is called.
+
+### Design of `CurlP`
+
+`CurlP` is designed as a hash function that acts on a `T1B1` binary-encoded ternary buffer, with a hash length of `243`
```
Alex6323: consistency: in line 76 it's *binary-coded*, here it's called *binary-encoded*

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+trits and an inner state of `729` trits, which we take as constants:
+
+```rust
+const HASH_LENGTH: usize = 243;
+const STATE_LENGTH: usize = HASH_LENGTH * 3;
+```
+
+In addition, a lookup table is used as part of the absorption step.
+2 is used to pad the table since there are only 9 combinations should be taken:
```
Alex6323: the explanation is probably a little bit confusing to the reader. The reason for the pad is that the lookup function is undefined for {3, 7}, and it allows us to not explicitly handle those edge cases.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+        Self(CurlP::new(CurlPRounds::Rounds81))
+    }
+}
+```
+
+### Design of `Kerl`
+
+The actual cryptographic hash function underlying `Kerl` is `keccak-384`. The actual task here is to transform an input
+of 243 (balanced) trits to 384 bits in a correct and performant way. This is done by interpreting the 243 trits as a
+signed integer `I` and converting it to a binary basis. In other words, the ternary coded integer is expressed as the
```
Alex6323: *ternary coded*? consistency please :smile:

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+to base `2`, i.e. find a a series such that:
+
+```
+I = b_0 * 2^0 + b_1 * 2^1 + ... + b_382 * 2^382 + b_383 * 2^383,
+```
+
+with `b_i` the bit at the `i-th` position.
+
+Assuming there exists an implementation of `keccak`, the main work in implementing `Kerl` is writing an efficient
+converter betWe en the ternary array interpreted as an integer, and its binary representation. For the binary
```
Alex6323: typo *between*

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+representation, one can either use an existing big integer library, or write one from scratch with only a subset of
```
Alex6323: *big-integer* or *Big Integer* library

text/0034-ternary-hash.md
thibault-martinez: Which one do you prefer ?

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+Remembering that each ternary array is actually some integer `I`, this is akin to adding another integer `H` to it, with
+all trits in the ternary buffer representing it set to `1`, where `I'` is `I` shifted into unsigned space, and the
+underscore `_t` means a ternary representation (either balanced or unbalanced):
+
+```
+I_t + H_t = [0, -1, +1, -1, -1, 0, 0, 0, +1] + [+1, +1, +1, +1, +1, +1, +1, +1] = I'_t
+```
+
+After changing the base to binary using some function which we call `to_bin` and which we require to distribute over
+addition (because the base in which a number is represented should have no bearing over adding said numbers), `H` needs
```
Alex6323: *should have* sounds a bit insecure. I would state this as a fact.

text/0034-ternary-hash.md
Alex6323: To me the whole term is confusing, but I know that it's being used in ternary science, so we need to use it. Then I think I'ld prefer *binary-encoded*. :slightly_smiling_face:

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,415 @@
...
+
+with `b_i` the bit at the `i-th` position.
+
+Assuming there exists an implementation of `keccak`, the main work in implementing `Kerl` is writing an efficient
+converter betWe en the ternary array interpreted as an integer, and its binary representation. For the binary
+representation, one can either use an existing big integer library, or write one from scratch with only a subset of
+required methods to make the conversion work.
+
+#### Important implementation details
+
```
Alex6323, line 310: I would describe the whole algo in a bullet list before diving into the details (if that's even necessary):
Input: `I_bt` [BTrit; 243] **b**alanced **t**ernary trit array
* Shift `I_bt` into **u**nbalanced **t**ernary space to receive `I'_ut` by adding the constant `H` [+1; 243] index-wise ,
* Calculate the **d**ecimal big-integer `I'_d` from `I'_ut` by setting those as coefficients in the number series evaluation for base=3
* Determine the binary representation `bin(I'_d)` of `I'_d`,
* Undo the shift by calculating: `bin(I_d) = bin(I'_d) - bin(H)` (represented now as [u8; 48] ??)
* Apply `keccak384(bin(I_d))` ??

Is this the correct process? I wanted to better understand it myself, hence this comment.

text/0034-ternary-hash.md
thibault-martinez: Now they do.

text/0034-ternary-hash.md
thibault-martinez: I don't really think this is an extension. These are core features of the Sponge. If we had to do that, everything needed a sponge would actually need to depend on both Sponge and SpongeExt, that would be a bit cumbersome IMO.

text/0034-ternary-hash.md
thibault-martinez: Clarified it.

text/0034-ternary-hash.md
thibault-martinez: I changed the order by putting an overview of the whole process first.

zesterer (dismissed):

samuel-rufi (dismissed):

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+    /// An error indicating that a failure has occured during a sponge operation.
+    type Error;
+
+    /// Absorb `input` into the sponge.
+    fn absorb(&mut self, input: &Trits) -> Result<(), Self::Error>;
+
+    /// Reset the inner state of the sponge.
+    fn reset(&mut self);
+
+    /// Squeeze the sponge into a buffer
```
Alex6323: punctuation :trollface:

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+    fn squeeze_into(&mut self, buf: &mut Trits) -> Result<(), Self::Error>;
+
+    /// Convenience function using `Sponge::squeeze_into` to return an owned version of the hash.
+    fn squeeze(&mut self) -> Result<TritBuf, Self::Error> {
+        let mut output = TritBuf::zeros(HASH_LENGTH);
+        self.squeeze_into(&mut output)?;
+        Ok(output)
+    }
+
+    /// Convenience function to absorb `input`, squeeze the sponge into a buffer, and reset the sponge in one go.
```
Alex6323: ..... the sponge into a buffer `buf`, .... (for consistency)

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+    fn digest_into(&mut self, input: &Trits, buf: &mut Trits) -> Result<(), Self::Error> {
+        self.absorb(input)?;
+        self.squeeze_into(buf)?;
+        self.reset();
+        Ok(())
+    }
+
+    /// Convenience function to absorb `input`, squeeze the sponge, and reset the sponge in one go.
+    /// Returns an owned version of the hash.
```
Alex6323, line 119: "Returns an owned version of the hash" (the actual return type is `TritBuf` though, so maybe this requires a little explanation like: "Returns an owned version of the hash stored in a TritBuf". Wdyt?

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+        let output = self.squeeze()?;
+        self.reset();
+        Ok(output)
+    }
+}
+```
+
+Following the sponge metaphor, an input provided by the user is `absorb`ed, and an output will be `squeeze`d from the
+data structure. `digest` is a convenience method calling `absorb` and `squeeze` in one go. The `*_into` versions of
+these methods are for passing a buffer into which the calculated hashes are written. The internal state will not be
```
Alex6323: "...which the calculated hashes are written **to**."

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+cleared unless `reset` is called.
+
+### Design of `CurlP`
+
+`CurlP` is designed as a hash function that acts on a `T1B1` binary-encoded ternary buffer, with a hash length of `243`
```
Alex6323: I would add a comma behind *hash function*, so that it's clear that ", with a hash length of ...", continues the main sentence.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+|**0**|0|-1|1|
+|**1**|-1|0|0|
+
+Given two balanced trits `t` and `t'`, the following table can easily be accessed by shifting them to unbalanced trits:
+```rust
+const TRUTH_TABLE: [[i8; 3]; 3] = [[1, 0, -1], [1, -1, 0], [-1, 1, 0]];
+```
+
+The way `CurlP` is defined, it can not actually fail, because the input or outputs can be of arbitrary size; hence,
+the associated type `Error = Infallible`.
```
Alex6323, line 157: I missed the line where the RFC clearly states, that a block `impl Sponge for CurlP` exists. It does provide it for the blanket impl, so I think it would be consistent as part of the "detailed design" to at least mention those impl blocks for CurlP and Kerl as well.

text/0034-ternary-hash.md
thibault-martinez: Hmm not so sure about that, there is already `into`.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        &mut self.0
+    }
+}
+```
+
+This allows using them as a `Sponge` as well if there is a blanket implementation like this:
+
+```rust
+impl<T: Sponge, U: DerefMut<Target = T>> Sponge for U {
```
Alex6323, line 247: I mentioned it in a comment before: this blanket impl would be more logical, if the reader had seen a `impl Sponge for CurlP` before.

text/0034-ternary-hash.md
thibault-martinez: Sorry I am not sure to understand this comment. And fail to see how it relates to the code you selected.

thibault-martinez: Hmm I now understand given your next comment but tbh I don't really know what to think about it. It's obvious impl will exist. I'm not sure the value to add them in the RFC since I'd definitely not impl the methods so it's basically writing the trait two more times.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+```
+
+# Drawbacks
+
++ All hash functions, no matter if they can fail or not, have to implement `Error`;
++ There are a lot of types. Is it important to encode that the most significant trit is `0` by having a `T242`?;
+
+# Rationale and alternatives
+
++ `CurlP` and `Kerl` are fundamental to the `iri v1.8.1` mainnet. They are thus essential for compatibility with it;
```
Alex6323: This is not up-to-date anymore and might be confusing as Bee is not intended anymore to be compatible with that IRI. The arguments are still valid, though.

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,472 @@
...
+U384<LittleEndian, U32Repr>::add_digit_inplace
+U384<LittleEndian, U32Repr>::sub_inplace
+U384<LittleEndian, U32Repr>::from_t242
+U384<LittleEndian, U32Repr>::try_from_t243
+```
+
+# Drawbacks
+
++ All hash functions, no matter if they can fail or not, have to implement `Error`;
++ There are a lot of types. Is it important to encode that the most significant trit is `0` by having a `T242`?;
```
Alex6323: regarding: "is it important to encode...." Isn't that an "unresolved question"? IMHO: "there are a lot of types" is not in-and-of-itself a drawback if the design justifies each type.

text/0034-ternary-hash.md
thibault-martinez: But this RFC is not about Bee as a whole, it's only the crypto crate and it's true that this crate is fundamental to the `iri v1.8.1` mainnet. Do you want to bump the version ?

text/0034-ternary-hash.md
thibault-martinez: What would be the returned type if it wasn't `TritBuf` though ?

text/0034-ternary-hash.md
Alex6323: ahh, sorry, my bad

Alex6323: I mistakenly assigned the `into` to the verb `passing` like "...are for passing a buffer into, which the calculated hashes are written to". So a comma would have helped me. But it's probably okay as it is. Don't know :smile:

text/0034-ternary-hash.md
Alex6323: well, since this RFC introduces `Hash`, and also introduces hash functions, that produce hashes, I would expect the return type to be `Hash`, and not `TritBuf`.

text/0034-ternary-hash.md
Alex6323: ^^^ updated

text/0034-ternary-hash.md
thibault-martinez: Ahh, but it can't be `Hash`, the output of the sponge is of variable size.

text/0034-ternary-hash.md
thibault-martinez: But there are still both `into` and `to` in your version.

text/0034-ternary-hash.md
Alex6323: yea, I knew that response would be coming (for sure it's trivial that the `Sponge` trait would be implemented on them), and I think intelligent readers will draw this conclusion immediatedly. It's your choice. Just wanted to mention it, that it *might* confuse some readers.

text/0034-ternary-hash.md
Alex6323: Idk whats best. Bumping it to the last version of IRI (1.8.6, right?) seems to be a good idea.

thibault-martinez: Sure 👍

text/0034-ternary-hash.md
Alex6323: right

Alex6323: but theoretically `digest` could return `Hash`.

text/0034-ternary-hash.md
Alex6323: yea, it's just the missing comma that got me confused. The RFC sentence makes sense now. My mistake was to put (in my mind) a comma behind `into`, and not already behind `buffer`.

text/0034-ternary-hash.md
thibault-martinez: Yes. Up until recently there was still this `OUT_LEN` that would have totally prevented doing that. Even if this is technically possible to return an actual Hash, I wonder if we should do it, the API would look weird with different types idk.

text/0034-ternary-hash.md
thibault-martinez: I can add that comma :)

text/0034-ternary-hash.md
Alex6323: don't know either. definitly not an `approval` stopper, but we should maybe keep this in mind if we want to change that. I'ld personally like to have `Hash` returned for `digest` as it's convenient (I think), logical, and gives the introduction of `Hash` a usecase already.

thibault-martinez: I tried to change but given the current definition of `Hash` this would mean a lot of new copies, since this is a very hot path I don't think that's wise to change :(

text/0034-ternary-hash.md
thibault-martinez: Sure

text/0034-ternary-hash.md
Alex6323: alrighty :+1: Thanks for trying!!!

text/0034-ternary-hash.md
thibault-martinez: Added

Alex6323 (dismissed):

samuel-rufi (dismissed):

text/0034-ternary-hash.md
```diff
@@ -0,0 +1,498 @@
...
+```
+
+```rust
+impl Sponge for CurlP {
+    type Error = Infallible;
+    ...
+}
+```
+
+In addition, the two wrapper types for the very common `CurlP` variants with `27` and `81` rounds:
```
zesterer: Is it possible to explain a little more about the justification for these wrapper types?

zesterer (approved):

samuel-rufi (approved):

Alex6323 (approved):

PR #36, opened 2020-05-28
RFC: `ternary`

zesterer:
Probably needs a few more examples: taking recommendations!

[Rendered](https://github.com/zesterer/bee-rfcs/blob/ternary/text/0036-ternary.md)

### DIFF
--- a/text/0036-ternary.md
+++ b/text/0036-ternary.md
@@ -1,0 +1,280 @@
++ Feature name: `ternary`
++ Start date: 2020-05-28
++ RFC PR: [iotaledger/bee-rfcs#36](https://github.com/iotaledger/bee-rfcs/pull/36)
++ Bee issue: [iotaledger/bee#80](https://github.com/iotaledger/bee/issues/80)
+
+# Summary
+
+This RFC introduces both the API and implementation of the `bee-ternary`
+crate, a general-purpose ternary manipulation crate written in Rust that may be
+used to handle ternary data in the IOTA ecosystem.
+
+# Motivation
+
+Ternary has been a fundamental part of the IOTA technology and is used
+throughout, including for fundamental features like wallet addresses and transaction
+identification.
+
+More information about ternary in IOTA can be found
+[here](https://docs.iota.org/docs/getting-started/0.1/introduction/ternary).
+
+*Note that the IOTA foundation is seeking to reduce the prevalence of ternary
+in the IOTA protocol to better accomodate modern binary architectures. See
+[Chrysalis](https://blog.iota.org/chrysalis-b9906ec9d2de) for more information.*
+
+Manipulating ternary in an ergonomic manner on top of existing binary-driven
+code can require complexity and, up until now, comes either with a lot of
+boilerplate code, implementations that are difficult to verify as correct, or a
+lack of features.
+
+`bee-ternary` seeks to become the canonical ternary crate in the Rust IOTA
+ecosystem by providing an API that is efficient, ergonomic, and featureful all
+at once.
+
+`bee-ternary` will allow IOTA developers to more easily write code that works
+with ternary and allow the simplification of existing codebases through use of
+the API.
+
+Broadly, there are 3 main benefits that `bee-ternary` aims to provide
+over use-specific implementations of ternary manipulation code.
+
+- **Ergonomics**: The API should be trivial to use correctly and should
+naturally guide developers towards correct and efficient solutions.
+
+- **Features**: The API should provide a fundamental set of core features that
+can be used together to cover most developer requirements. Examples of such
+features include multiple encoding schemes, binary/ternary conversion,
+(de)serialization, and tryte string de/encoding.
+
+- **Performance**: The API should allow 'acceptably' efficient manipulation of
+ternary data. This is difficult to ruggedly define, and implementation details
+may change later, but broadly this RFC intends to introduce an API that does not
+inherently inhibit high performance through poor design choices.
+
+# Detailed design
+
+`bee-ternary` is designed to be extensible and is built on top of a handful of
+fundamental abstraction types. Below are listed the features of the API and a
+summary of the core API features.
+
+## Features
+
+- Efficient manipulation of ternary buffers (trits and trytes).
+- Multiple encoding schemes.
+- Extensible design that allows it to sit on top of existing data structures,
+avoiding unnecessary allocation and copying.
+- An array of utility functions to allow for easy manipulation of ternary data.
+- Zero-cost conversion between trit and tryte formats (i.e: no slower than
+the equivalent code would be if hand-written).
+
+## Key Types & Traits
+
+The crate supports both balanced and unbalanced trit representations, supported
+through two types, `Btrit` and `Utrit`. `Btrit` is the more common
+representation and so is the implicit default for most operations in the crate.
+
+### `Btrit` and `Utrit`
+
+```rust
+enum Btrit { NegOne, Zero, PlusOne }
+enum Utrit { Zero, One, Two }
+```
+
+### `Trits`
+
+```rust
+struct Trits<T: RawEncoding> { ... }
+```
+
+- Generic across different trit encoding schemes (see below).
+- Analogous to `str` or `[T]`.
+- Unsized type, represents a buffer of trits of a specified encoding.
+- Most commonly used from behind a reference (as with `&str` and `&[T]`),
+representing a 'trit slice'.
+- Can be created from a variety of types such as`TritBuf` and `[i8]` with
+minimal overhead.
+- Sub-slices can be created from trit slices at a per-trit level.
+
+### `TritBuf`
+
+```rust
+struct TritBuf<T: RawEncodingBuf> { ... }
+```
+
+- Generic across different trit encoding schemes (see below).
+- Analogous to `String` or `Vec<T>`.
+- Most common way to manipulate trits.
+- Allows pushing, popping, and collecting trits.
+- Implements `Deref<Target=Trits>` (in the same way that `Vec<T>` implements
+`Deref<[T]>`).
+
+### `RawEncoding`
+
+```rust
+trait RawEncoding { ... }
+```
+
+- Represents a raw trit buffer of some encoding.
+- Common interface implemented by all trit encodings (`T1B1`, `T3B1`, `T5B1`,
+etc.).
+- Largely an implementation detail, not something you need to care about unless
+implementing your own encodings.
+- Minimal implementation requirements: safety and most utility functionality is
+provided by `Trits` instead.
+
+### `RawEncodingBuf`
+
+```rust
+trait RawEncodingBuf { ... }
+```
+
+- Buffer counterpart of `RawEncoding`, always associated with a specific
+`RawEncoding`.
+- Distinct from `RawEncoding` to permit different buffer-like data structures
+in the future (linked lists, stack-allocated arrays, etc.).
+
+### `T1B1`/`T2B1`/`T3B1`/`T4B1`/`T5B1`
+
+```rust
+struct TXB1 { ... }
+```
+
+- Types that implement `RawEncoding`.
+- Allow different encodings in `Trits` and `TritBuf` types.
+
+### `T1B1Buf`/`T2B1Buf`/`T3B1Buf`/`T4B1Buf`/`T5B1Buf`
+
+```rust
+struct TXB1Buf { ... }
+```
+
+- Types that implement `RawEncodingBuf`.
+- Allow different encodings in `TritBuf` types.
+- Each type is associated with a `RawEncoding` type.
+
+## API
+
+The API makes up the body of this RFC. Due to its considerable length, this RFC
+simply refers to the documentation in question. You can find those docs in the
+`bee` repository.
+
+## Encodings
+
+`bee-ternary` supports many different trit encodings. Notable encodings are
+explained in the crate documentation.
+
+## Common Patterns
+
+When using the API, the most common types interacted with are `Trits` and
+`TritBuf`. These types are designed to play well with the rest of the Rust
+ecosystem. Here follows some examples of common patterns that you may wish to
+make use of.
+
+### Turning some `i8`s into a trit buffer
+
+```rust
+[-1, 0, 1, 1, -1, 0]
+	.iter()
+	.map(|x| Btrit::try_from(*x).unwrap())
+	.collect::<TritBuf>()
+```
+
+Alternatively, for the `T1B1` encoding only, you may directly reinterpret the
+`i8` slice.
+
+```rust
+let xs = [-1, 0, 1, 1, -1, 0];
+
+Trits::try_from_raw(&xs, xs.len())
+	.unwrap()
+```
+
+If you are *certain* that the slice contains *only* valid trits then it is
+possible to unsafely reinterpret with amortised O(1) cost (i.e: it's basically
+free). However, scenarios in which this is necessary and sound are exceedingly
+uncommon. If you find yourself doing this, ask yourself first whether it is
+necessary.
+
+```rust
+let xs = [-1, 0, 1, 1, -1, 0];
+
+unsafe { Trits::from_raw_unchecked(&xs, xs.len()) }
+```
+
+### Turning a trit slice into a tryte string
+
+```rust
+trits
+	.iter_trytes()
+	.map(|trit| char::from(trit))
+	.collect::<String>()
+```
+
+This becomes even more efficient easier with a `T3B1` trit slice since it has
+the same underlying representation as a tryte slice.
+
+```rust
+trits
+	.as_trytes()
+	.iter()
+	.map(|trit| char::from(*trit))
+	.collect::<String>()
+```
+
+### Turning a tryte string into a `T3B1` trit buffer
+
+```rust
+tryte_str
+	.chars()
+	.map(Tryte::try_from)
+	.collect::<Result<TryteBuf, _>>()
+	.unwrap()
+```
+
+Since this is a common operation, there exists a shorthand.
+
+```rust
+TryteBuf::try_from_str(tryte_str).unwrap()
+```
+
+### Turning a trit slice into a trit buffer
+
+```rust
+trits.to_buf()
+```
+
+### Turning a trit slice into a trit buffer of a different encoding
+
+```rust
+trits.encode::<T5B1>()
+```
+
+### Overwriting a sub-slice of a trit with copies of a trit
+
+```rust
+trits[start..end].fill(Btrit::Zero)
+```
+
+### Copying trits from a source slice to a destination slice
+
+```rust
+tgt.copy_from(&src[start..end])
+```
+
+# Drawbacks
+
+This RFC does not have any particular drawbacks over an alternative approach.
+Ternary is currently essential to the function of much of the IOTA ecosystem
+and Bee requires a ternary API of some sort in order to effectively operate
+within it.
+
+# Rationale and alternatives
+
+No suitable alternatives exist to this RFC. Rust does not have a mature ternary
+manipulation crate immediately available that suits our needs.
+
+# Unresolved questions
+
+The API has now been in use in Bee for several months. Questions about
+additional API features exist, but the current API seems to have proved its
+suitability.


text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
++ Feature name: (fill me in with a unique ident, `my_awesome_feature`)
++ Start date: (fill me in with today's date, YYYY-MM-DD)
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
```
Alex6323: You can fill this section in now. I added the corresponding tracking issue: [iotaledger/bee#80](https://github.com/iotaledger/bee/issues/80)

thibault-martinez: Also don't forget to rename the file!

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+# Summary
+
+In this RFC we introduce both the API and implementation of the `bee-ternary`
+crate, a general-purpose ternary manipulation crate written in Rust that may be
+used to handle ternary data in the IOTA ecosystem.
+
+# Motivation
+
+Ternary is a fundamental part of the IOTA ecosystem and is used throughout, even
+for fundamental features like wallet addresses and transaction identification.
```
Alex6323: "...of the IOTA technology"

Alex6323: That statement is true, but I wonder whether we want to emphasize on the "fundamental" part. Can we even write something like: "Ternary **has been** a fundamental part of the IOTA technology and is still used throughout today. This includes address generation, hashing, and signature verification, as well as internal transaction representation and proof-of-work."
Then maybe add something like a warning/disclaimer, maybe something like:
"Note, that the IOTA Foundation is attempting to reduce the ternary aspects of the protocol to better accommodate  for todays binary architectures [see](https://blog.iota.org/chrysalis-b9906ec9d2de), hence, this crate will only be needed until a full transition to a new signing scheme."

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+
+`bee-ternary` seeks to become the canonical ternary crate in the Rust IOTA
+ecosystem by providing an API that is efficient, ergonomic, and featureful all
+at once.
+
+We hope that `bee-ternary` will allow IOTA developers to more easily write code
+that works with ternary and allow the simplification of existing codebases
+through use of the API.
+
+Broadly, there are 3 main benefits things that `bee-ternary` aims to provide
```
Alex6323: ~~things~~

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
++ Feature name: (fill me in with a unique ident, `my_awesome_feature`)
++ Start date: (fill me in with today's date, YYYY-MM-DD)
++ RFC PR: [iotaledger/bee-rfcs#0000](https://github.com/iotaledger/bee-rfcs/pull/0000)
++ Bee issue: [iotaledger/bee#0000](https://github.com/iotaledger/bee/issues/0000)
+
+# Summary
+
+In this RFC we introduce both the API and implementation of the `bee-ternary`
```
Alex6323: I am in favor of using a more *neutral* form like "This RFC introduces/proposes..." rather then first person pronouns.

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+
+- **Performance**: The API should allow 'acceptably' efficient manipulation of
+ternary data. This is difficult to ruggedly define, and implementation details
+may change later, but broadly we intend to introduce an API that doesn't
+inherently inhibit high performance through poor design choices.
+
+# Detailed design
+
+`bee-ternary` is designed to be extensible and is built on top of a handful of
+powerful abstraction types. Below we list the features of the API and summarise
```
Alex6323: summari**ze** :trollface: 
(sorry)

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+a few of the core API objects.
```
Alex6323: Can you expand on why you talk about *objects* instead of *types*. I you don't want to be too explicit here, I would prefer the term *features*.

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Features
+
+- Efficient manipulation of ternary buffers (trits and trytes)
+- Multiple encoding schemes
+- Extensible design that allows it to sit on top of existing data structures,
+avoiding unnecessary allocation and copying
```
Alex6323: missing bullet point/dash

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+- A plethora of utility functions to allow for easy manipulation of ternary data
```
Alex6323: you use terms like *powerful* and *plethora* which sounds a bit like a pitch (just my impression, I might be wrong). I know it's boring, but I would prefer a more formal, descriptive wording. Note that this would be just **my** preference. I am fine with as it is, if the others think it's fine :smile:

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+- Zero-cost conversion between trit and tryte formats
```
Alex6323: just making sure it's correct what we state here: Is it really 0-cost? If an actual conversion is happening, then I wonder how that can be 0-cost. But if you say that it's just a *reinterpretation* of the underlying bits, then it's fine. But isn't the byte encoding of a trit different from the byte encoding of a tryte, so how can this be 0-cost?

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+```rust
+struct Trits<T: RawEncoding> { ... }
+```
+
+- Analogous to `str` or `[T]`
+- Unsized type, represents a buffer of trits of a specified encoding
+- Most commonly used from behind a reference (as with `&str` and `&[T]`),
+representing a 'trit slice'
+- Can be created from a variety of types such as`TritBuf` and `[i8]` with
+minimal overhead
```
Alex6323: missing bullet points/dashes

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+avoiding unnecessary allocation and copying
+- A plethora of utility functions to allow for easy manipulation of ternary data
+- Zero-cost conversion between trit and tryte formats
+
+## Key Types & Traits
+
+### `Trits`
+
+```rust
+struct Trits<T: RawEncoding> { ... }
```
Alex6323: Maybe we should move the `RawEncoding` trait before this `struct`? Then it might be easier to understand why `Trits` generalizes over it. What do you think?

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+`TritBuf`. These types are designed to play well with the rest of the Rust
+ecosystem. Here follows some examples of common patterns that you may wish to
+make use of.
+
+### Turning some `i8`s into a trit buffer
+
+```rust
+[-1, 0, 1, 1, -1, 0]
+	.iter()
+	.map(|x| Btrit::try_from(*x).unwrap())
```
Alex6323: `Btrit` probably needs to be mentioned earlier.

text/0000-ternary.md
```diff
@@ -0,0 +1,238 @@
...
+- Most commonly used from behind a reference (as with `&str` and `&[T]`),
+representing a 'trit slice'
+- Can be created from a variety of types such as`TritBuf` and `[i8]` with
+minimal overhead
+- Sub-slices can be created from trit slices at a per-trit level
+
+### `TritBuf`
+
+```rust
+struct TritBuf<T: RawEncodingBuf> { ... }
```
Alex6323: I think we should also provide the struct and trait (down below) members? What do you think?

text/0000-ternary.md
zesterer: I use the term to talk about distinct API things, but this could include traits. I'll use the term 'features' if you wish.

text/0000-ternary.md
zesterer: This is intentional, they're all the same point

text/0000-ternary.md
zesterer: It's a reinterpretation in the case of `T3B1`, yes.

Regardless, the canonical usage of 'zero-cost' doesn't mean "requires not computation" - it just means that there's no cost imposed over a manual implementation of the feature (see https://idursun.com/posts/rust_zero_cost_abstractions_in_action/).

text/0000-ternary.md
zesterer: Intentional again

text/0000-ternary.md
zesterer: Will do

text/0000-ternary.md
zesterer: There's a lot of documentation that would come with that, and we'd need to provide many impl blocks (since methods differ for different type parameters). This would be virtually equivalent to pasting most of the codebase into the RFC.

Also, the internal field(s) are an implementation detail and should not be exposed to the API user.

text/0000-ternary.md
zesterer: Changed :)

text/0036-ternary.md
```diff
@@ -0,0 +1,254 @@
++ Feature name: ternary
```
thibault-martinez:
```suggestion
+ Feature name: `ternary`
```
For consistency with the other RFCs :p

text/0000-ternary.md
zesterer: I'm against this largely because it's not a common thing for an API user to interact with. As of yet, no code in the rest of Bee explicitly uses this API feature and so I'm reluctant to put it first since it may confuse users.

text/0000-ternary.md
Alex6323: Not bothered so much tbh :smile: . If you think it's fine, then go for it. My problem with it is that it has different semantics attached to it depending on the context, but essentially it's really just a minor thing. You can close this conversation, if you're sure, that it won't cause confusion (alternatives: features, models, concepts)

text/0000-ternary.md
Alex6323: ah, yea, thanks for the reminder. I still find it a bit confusing (and potentially some readers as well?). Maybe add this link to help readers understand this Rust notion of 0-cost?

thibault-martinez (commented):
A tiny nit but something I have been told by the research department, we usually try to avoid contractions (like `doesn't`) in RFCs.

text/0000-ternary.md
Alex6323: okay. I understand. Maybe an introductory sentence then. Because atm when the reader reaches this point of the RFC (and understood everything so far), he is now confronted with a `struct` definition that uses a still undefined trait bound `RawEncoding`.

text/0000-ternary.md
Alex6323: hmmm.... This RFC is supposed to guide the implementer (which in general is a different person than the RFC author). It's not written for the API user. I think there is no need to provide `impl` blocks, but shouldn't the implementer know, how this type is supposed to be realized internally (hence field definitions)?

text/0036-ternary.md
```diff
@@ -0,0 +1,254 @@
...
+# Summary
+
+This RFC introduces both the API and implementation of the `bee-ternary`
+crate, a general-purpose ternary manipulation crate written in Rust that may be
+used to handle ternary data in the IOTA ecosystem.
+
+# Motivation
+
+Ternary has been a fundamental part of the IOTA technology and is used
+throughout, even for fundamental features like wallet addresses and transaction
```
thibault-martinez: Is it just me or this `even` sounds a bit negative ?

text/0036-ternary.md
```diff
@@ -0,0 +1,254 @@
...
++ Start date: 2020-05-28
++ RFC PR: [iotaledger/bee-rfcs#36](https://github.com/iotaledger/bee-rfcs/pull/36)
++ Bee issue: [iotaledger/bee#80](https://github.com/iotaledger/bee/issues/80)
+
+# Summary
+
+This RFC introduces both the API and implementation of the `bee-ternary`
+crate, a general-purpose ternary manipulation crate written in Rust that may be
+used to handle ternary data in the IOTA ecosystem.
+
```
thibault-martinez, line 11: I'd add a link to I'd add a link to https://docs.iota.org/docs/getting-started/0.1/introduction/ternary

text/0036-ternary.md
```diff
@@ -0,0 +1,254 @@
...
+`Deref<[T]>`)
+
+### `RawEncoding`
+
+```rust
+trait RawEncoding { ... }
+```
+
+- Represents a raw trit buffer of some encoding
+- Common interface implemented by all trit encodings (`T1B1`, `T3B1`, `T5B1`,
```
thibault-martinez, line 118: These are never actually defined, should we expand the `Encoding` a tiny bit and define these acronyms ?

text/0036-ternary.md
```diff
@@ -0,0 +1,254 @@
...
+- Buffer counterpart of `RawEncoding`, always associated with a specific
+`RawEncoding`
+- Distinct from `RawEncoding` to permit different buffer-like data structures
+in the future (linked lists, stack-allocated arrays, etc.)
+
+## API
+
+The API makes up the body of this RFC. Due to its considerable length, this RFC
+simply refers to the documentation in question. You can find those docs in the
+`bee-p` (and later `bee`) repository.
```
thibault-martinez: I would not even mention `bee-p` here. This would be false for some days but it's fine. Better than having this stick to the RFC IMO.

text/0036-ternary.md
Alex6323: No, doesn't sound negative. The point Josh is making here is that ternary is central to IOTA tech down to its core functionalities.

Alex6323: But I would rephrase this sentence to the one I suggested :stuck_out_tongue:

text/0036-ternary.md
zesterer: Will do

text/0036-ternary.md
```diff
@@ -0,0 +1,280 @@
...
+unsafe { Trits::from_raw_unchecked(&xs, xs.len()) }
+```
+
+### Turning a trit slice into a tryte string
+
+```rust
+trits
+	.chunks(3)
+	.map(|trits| char::from(Tryte::from_trits([
+		trits.get(0),
```
samuel-rufi: Does .get(x) still need the unwrap() or was the implementation changed?

text/0036-ternary.md
```diff
@@ -0,0 +1,280 @@
...
+Manipulating ternary in an ergonomic manner on top of existing binary-driven
+code can require complexity and, up until now, comes either with a lot of
+boilerplate code, implementations that are difficult to verify as correct, or a
+lack of features.
+
+`bee-ternary` seeks to become the canonical ternary crate in the Rust IOTA
+ecosystem by providing an API that is efficient, ergonomic, and featureful all
+at once.
+
+We hope that `bee-ternary` will allow IOTA developers to more easily write code
```
Alex6323: still a few **we**s used

text/0036-ternary.md
```diff
@@ -0,0 +1,280 @@
...
+- Zero-cost conversion between trit and tryte formats
+
+## Key Types & Traits
+
+### `Btrit` and `Utrit`
+
+```rust
+enum Btrit { NegOne, Zero, PlusOne }
+enum Utrit { Zero, One, Two }
+```
```
Alex6323, line 81: Please move this below the paragraph starting at line 78. Rather than providing code first, and then explain, I am in favor of leading the reader towards the "idea", then provide a code snippet, and then add possible notes/considerations/references.

text/0036-ternary.md
```diff
@@ -0,0 +1,280 @@
...
+
+The crate supports both balanced and unbalanced trit representations, supported
+through two types, `Btrit` and `Utrit`. `Btrit` is the more common
+representation and so is the implicit default for most operations in the crate.
+
+### `Trits`
+
```
Alex6323, line 84: Maybe it's just me but I wished there were a short introductory sentence for each sub-section before showing code. I do not wish to bloat the RFC. I like a minimalistic approach, but still....a short paragraph explaining the reasoning would help with the understanding why this RFCs suggests a certain abstraction.

text/0036-ternary.md
zesterer: Whoops, I think this was accidental

text/0036-ternary.md
zesterer: I'm not sure what information I could provide here. Do you have any specific ideas?

text/0036-ternary.md
zesterer: Yes, I'm just being very stupid today :smile:

text/0036-ternary.md
```diff
@@ -0,0 +1,280 @@
...
+### `T1B1Buf`/`T2B1Buf`/`T3B1Buf`/`T4B1Buf`/`T5B1Buf`
+
+```rust
+struct TXB1Buf { ... }
+```
+
+- Types that implement `RawEncodingBuf`
+- Allow different encodings in `TritBuf` types.
+- Each type is associated with a `RawEncoding` type.
```
thibault-martinez, line 153: Tiny nit again, sorry about my OCD triggering, you sometimes end list items with a dot or not

zesterer: Will fix this

text/0000-ternary.md
zesterer: Will address this with an explanation

text/0000-ternary.md
zesterer: Will add an explanation

text/0000-ternary.md
zesterer: Have changed it now

text/0000-ternary.md
zesterer: Those field definitions might well change later on. Besides, the fields don't actually relate to the implementation all that much since they rely on unsized type tricks to work. I really don't think it would be useful to do more than pointing to the actual implementation given the complexities associated with it.

text/0036-ternary.md
zesterer: Have changed this

text/0000-ternary.md
Alex6323: Alright, that works for me. 

For the record: I wouldn't be too much concerned with your first statement, that field definitions might change. That's true, but I think RFCs are not supposed to be necessarily set in stone, and can be updated later on. If the changes are too significant, then an additional improvement or even replacement RFC should be the way to go. :smile:

Alex6323: Again...I think my concerns were addressed. Thanks.

Alex6323 (dismissed):
lgtm now!

text/0036-ternary.md
```diff
@@ -0,0 +1,280 @@
...
+used to handle ternary data in the IOTA ecosystem.
+
+# Motivation
+
+Ternary has been a fundamental part of the IOTA technology and is used
+throughout, including for fundamental features like wallet addresses and transaction
+identification.
+
+More information about ternary in IOTA can be found here:
+(https://docs.iota.org/docs/getting-started/0.1/introduction/ternary)
```
thibault-martinez: Are the () intentional ?

thibault-martinez (dismissed):

Alex6323 (dismissed):

samuel-rufi (dismissed):

text/0036-ternary.md
```diff
@@ -0,0 +1,280 @@
...
+
+- **Performance**: The API should allow 'acceptably' efficient manipulation of
+ternary data. This is difficult to ruggedly define, and implementation details
+may change later, but broadly this RFC intends to introduce an API that does not
+inherently inhibit high performance through poor design choices.
+
+# Detailed design
+
+`bee-ternary` is designed to be extensible and is built on top of a handful of
+fundamental abstraction types. Below are listed the features of the API and a
```
Alex6323, line 57: Not a native speaker, but "Below are listed the features...." sounds a bit off to me. Are you sure it's syntactically correct?

text/0036-ternary.md
```diff
@@ -0,0 +1,280 @@
...
+
+## Features
+
+- Efficient manipulation of ternary buffers (trits and trytes).
+- Multiple encoding schemes.
+- Extensible design that allows it to sit on top of existing data structures,
+avoiding unnecessary allocation and copying.
+- An array of utility functions to allow for easy manipulation of ternary data.
+- Zero-cost conversion between trit and tryte formats (i.e: no slower than
+the equivalent code would be if hand-written).
```
Alex6323, line 68: you probably gonna slaughter me, but I can't help to suggest using commas instead of periods for lists.

thibault-martinez: And I can't help but suggest using semicolon instead :p

text/0036-ternary.md
zesterer: Will change it. It's syntactically correct, but more cumbersome than it needs to be.

text/0036-ternary.md
zesterer: :upside_down_face:

text/0036-ternary.md
Alex6323: thanks

text/0036-ternary.md
Alex6323: lol, semicolon is also fine for me

samuel-rufi (approved):

thibault-martinez (approved):

Alex6323 (approved):

PR #39, opened 2020-06-29
RFC30: Remove erroneous mentions of broadcasting

thibault-martinez:
The message itself is not a broadcast at all, it's just a transaction. Broadcast is misleading in the sense we could think it's being sent to many peers, it's not.

### DIFF
--- a/text/0030-protocol-messages.md
+++ b/text/0030-protocol-messages.md
@@ -143,11 +143,11 @@ Up to 32 bytes are supported, limiting the number of protocol versions to 256. E
 
 Type ID: `2`
 
-A legacy message to broadcast a transaction and request another one at the same time.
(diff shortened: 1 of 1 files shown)


samuel-rufi (approved):

Alex6323:
can you please provide a description, why the term `broadcast` is wrong. Is it wrong in terms of the definition of the word, or is this about something else?

zesterer (approved):

thibault-martinez:
The message itself is not a broadcast at all, it's just a transaction. Broadcast is misleading in the sense we could think it's being sent to many peers, it's not.

Alex6323:
Ah, okay. That clears it up for me. May I suggest adding this explanation to the PR?

Alex6323 (approved):

PR #41, opened 2020-07-03
RFC: shutdown

Alex6323:
[Rendered](https://github.com/Alex6323/bee-rfcs/blob/rfc-shutdown/text/0041-shutdown.md)

### DIFF
--- a/text/0041-shutdown.md
+++ b/text/0041-shutdown.md
@@ -71,7 +71,6 @@ struct Shutdown {
 
 The API of `Shutdown` is very small. It only needs to allow filling those internal lists, and provide an `execute` method to facilitate the actual shutdown:
 
-The implementation of `Shutdown` looks like this:
 ```rust
 impl Shutdown {
 
--- a/text/0041-shutdown.md
+++ b/text/0041-shutdown.md
@@ -1,7 +1,7 @@
 + Feature name: `shutdown`
 + Start date: 2020-07-03
-+ RFC PR: [iotaledger/bee-rfcs#00xx](https://github.com/iotaledger/bee-rfcs/pull/00xx)
-+ Bee issue: [iotaledger/bee#00xx](https://github.com/iotaledger/bee/issues/00xx)
++ RFC PR: [iotaledger/bee-rfcs#41](https://github.com/iotaledger/bee-rfcs/pull/41)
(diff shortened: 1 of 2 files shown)


text/0041-shutdown.md
```diff
@@ -0,0 +1,135 @@
...
+// worker terminates, i.e. shuts down. To make sure, the shutdown is complete, one has
+// to store those `Future`s in a list, and await all of them iteratively before exiting
+// the program.
+type WorkerShutdown = Box<dyn Future<Output = Result<(), WorkerError>> + Unpin>;
+
+// Before shutting down a system, workers oftentimes need to perform one or more final
+// actions to prevent data loss and/or data corruption. Therefore each worker can register
+// an arbitrary amount of actions, that will be executed as the final step of the shutdown
+// process.
+type Action = Box<dyn Fn()>;
```
zesterer: Can this be extended to `FnOnce()`? This should be possible since the vector gets drained during the shutdown process.

samuel-rufi (dismissed):

text/0041-shutdown.md
Alex6323: I think so. :+1: Each action is supposed to be called only once.

zesterer (dismissed):

text/0041-shutdown.md
```diff
@@ -79,38 +79,38 @@ impl Shutdown {
...
-    }
-
     // Registers a worker shutdown future, which completes once the worker terminates in one
-    // way or another.
-    fn add_worker_shutdown(&mut self, worker: impl Future<Output = Result<(), WorkerError>> + Unpin + 'static) {
+    // way or another, and registers a shutdown notification channel by providing its sender half.
+    fn add_worker_shutdown(
+        &mut self,
+        worker: impl Future<Output = Result<(), WorkerError>> + Unpin + 'static,
+        notifier: ShutdownNotifier,
```
thibault-martinez, line 86: I think it would make more sense to reverse the params here. It makes sense because the notifier is created before the worker. That would also match the order in the struct, in the body of this func and in the body of execute.

Alex6323: okay 🙂

zesterer (approved):

thibault-martinez (approved):

samuel-rufi (approved):

PR #45, opened 2020-07-15
RFC: `api` 

samuel-rufi:
[Rendered](https://github.com/rufsam/bee-rfcs/blob/master/text/0045-api.md)

### DIFF
--- a/text/0045-api.md
+++ b/text/0045-api.md
@@ -14,7 +14,7 @@ With the help of this interface, applications can interact with the Bee node and
 
 # Detailed design
 
-The design describes a layered approach that provides services on which applications can be built.
+The design describes a layered approach that provides services on which applications can build.
 
 ## Service
 
@@ -100,7 +100,7 @@ impl From<TransactionsByBundleResponse> for JsonValue {
 ```
 
 Notable is the use of `HashItem` or `TransactionRefItem` in the conversions. These wrapper types are used to be conform with Rust's orphan rules. Since multiple services use for example the `Hash` type within parameters/response, conversions should be reusable.
-For this reason, conversions should be defined also for inner types (e.g. `Hash`) of a request/response.
+For this reason, `TryFrom/From`s should be defined also for inner types (e.g. `Hash`) of a request/response.
 ```
 pub struct HashItem(pub Hash);
 
--- a/text/0045-api.md
+++ b/text/0045-api.md
@@ -5,16 +5,15 @@
 
 # Summary
 
-This RFC proposes a crate that provides interfaces and conversions to support presentation layers such as web, mobile, or UI.
+This RFC proposes a crate that provides interfaces and tools to support presentation layers such as web, mobile, or UI. Therefore, the focus is on high-level interaction.
 
 # Motivation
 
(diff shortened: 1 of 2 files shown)


text/0045-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+# Detailed design
+
+The design describes a layered approach that provides services on which applications can build.
+
+## Service
+
+The `Service` trait defines all the services that a Bee node exposes. It hides internal implementation and changes, furthermore governs external access to data and functions. It can support multiple presentation layers such as web, mobile, UI, etc. 
+Since it hides internal logic and governs access, it helps maintain the layers above.
+The `Service` trait is defined as follows:
```
wusyong, line 23: Do this also mean topic subscription is included? I feel like this should split it to small section to fit the need of each individual service. Something like UserService, TopicService, QueryService etc.

text/0045-api.md
```diff
@@ -0,0 +1,191 @@
...
+}
+```
+
+## WebService trait
+
+The `WebService` trait specifies which endpoints a Web service should implement. This also means, it doesn't need to expose all functions of the `Service` trait.
+
+```
+#[async_trait]
+pub trait WebService {
```
wusyong, line 144: Extend from above. And maybe here we will have a meta trait like as a common service what this kind of service should have:

```suggestion
trait WebService: QueryService + UserService {
```

text/0045-api.md
```diff
@@ -0,0 +1,191 @@
...
+
+No drawbacks known to the author.
+
+# Rationale and alternatives
+
+No viable alternatives known to the author.
+
+# Unresolved questions
+
+- Should function params be created with a `new` constructor instead? And eventually return an error already there when illegal parameters where passed, instead in the service?
```
wusyong, line 190: I do feel like a builder pattern is more idiomatic in Rust. But that's personal taste if the interface is not going to be used by others (that is only bee devs and contributors need to build the layer by these traits.)

text/0045-api.md
samuel-rufi: Thats a really good question.
Unfortunately API is a broad term, it could mean low level API where libraries/plugins get loaded or high level API which supports web, UI etc. Therefore, one API might be totally different from another. Also, WebServices like REST, JSON-RPC don't have much in common as for example with MQTT.

For this reason, IMO we shouldn't mix up APIs. We probably won't be able to define 1 trait cleanly that represents all different kinds of APIs.
API's should be defined separately IMO and get their own interfaces/structure they need. Maybe it would work as you suggested, having different interfaces (`UserService, TopicService, QueryService`) that require other interfaces to be implemented (`trait WebService: UserService, TopicService`). However I'm not sure if this is the right way and if this flexiblity can even represent all these different kind of APIs. So IMO we should keep these different API's separate.

@thibault-martinez MQTT is currently developed in it's own crate (bee-mqtt). Should we move it into the `bee-api` folder? So we have an `bee-api/mqtt` folder and an `bee-api/web-service`.

text/0045-api.md
samuel-rufi: Good idea, I guess builder pattern would be nice here for parameters/response

zesterer:
Could you format with Rust like '```rust'?

PR #47, opened 2020-07-29
RFC:  `storage` 

louaykamel:
[Rendered](https://github.com/iotaledger/bee-rfcs/blob/bee-storage/text/0047-storage.md)

### DIFF
--- a/text/0047-storage.md
+++ b/text/0047-storage.md
@@ -6,8 +6,7 @@ tags: RFC, draft
 # bee-storage
 + Feature name: `storage`
 + Start date: 2020-07-15
-+ RFC PR: [iotaledger/bee-rfcs#46](https://github.com/iotaledger/bee-rfcs/pull/46)
-+ Bee issue:  [iotaledger/bee#106](https://github.com/iotaledger/bee/issues/106)
++ RFC PR: [iotaledger/bee-rfcs#47](https://github.com/iotaledger/bee-rfcs/pull/47)
 
 # Summary
 
--- a/text/0046-storage.md
+++ b/text/0046-storage.md
@@ -1,0 +1,218 @@
+---
+title: storage
+tags: RFC, draft
+---
+
+# bee-storage
++ Feature name: `storage`
++ Start date: 2020-07-15
++ RFC PR: [iotaledger/bee-rfcs#46](https://github.com/iotaledger/bee-rfcs/pull/46)
++ Bee issue:  [iotaledger/bee#106](https://github.com/iotaledger/bee/issues/106)
+
+# Summary
+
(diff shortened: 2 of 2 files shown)


text/0047-storage.md
```diff
@@ -0,0 +1,217 @@
...
+
+- Current Structures
+
+```rust
+/// The Storage struct will be created automatically by the assignment in feature
+/// Here we use rocks_db as an example for the storage structure
+
+#[cfg(feature = "rocks_db")]
+pub struct Storage {
+    inner: ::rocksdb::DB  // The storage backend
```
zesterer, line 63: Perhaps it should be the `inner` field that has the feature flag?

text/0047-storage.md
```diff
@@ -0,0 +1,217 @@
...
+    /// It takes the ownership of self, and returns ()
+    async fn shutdown(self) -> Result<(), Box<dyn Error>>;
+}
+#[cfg(feature = "rocks_db")]
+#[async_trait]
+impl Backend for Storage {
+    /// It starts RocksDB instance and then initialize the required column familes
+    async fn start(config_path: String) -> Result<Self, Box<dyn Error>> {
+        let config_as_string = fs::read_to_string(config_path)?;
+        let config: config::Config = toml::from_str(&config_as_string)?;
```
zesterer, line 120: Where is this `Config` type from? The `config` crate? If so, it might make more sense to have the config loading abstracted such that storage API users can customise it.

text/0047-storage.md
```diff
@@ -0,0 +1,217 @@
...
+                self.into_trits_allocated(tx_trit_buf.as_slice_mut());
+                let hash_buf = hash.to_inner().encode::<T5B1Buf>();
+                storage.inner.put_cf(
+                    &hash_to_tx,
+                    cast_slice(hash_buf.as_i8_slice()),
+                    cast_slice(tx_trit_buf.encode::<T5B1Buf>().as_i8_slice()),
+                )?;
+                Ok(())
+            }
+            async fn find_by_hash(hash: Hash, storage: &Storage) -> Result<Option<Self>, OpError> {
```
zesterer, line 164: `Hash` should be passed by reference to avoid unnecessary cloning (I think `Hash` implements `Deref<Trits>` now, so you can just call `.encode()` on it directly.

louaykamel: Perfect, yeah I will force reference everywhere.

text/0047-storage.md
louaykamel: Yes it's already abstracted using feature, you can check that in storage branch.

text/0047-storage.md
louaykamel: Interesting, I agree it makes more sense.

